PR #4, opened 2018-07-09
Updated threat IDs and added a few new threats

shambho:


izar (changes requested):
Needs getter/setters for new attributes.

PR #8, opened 2018-07-19
Added isSQL and storesSensitiveData to Datastore, corrected typo on i…

nozmore:
…nAWS and updated logging threat to use storesSensitiveData

-Added missing getters and setters for Datastore
-Corrected typo on Element, onAWS to inAWS
-Changed default value for storesSensitiveData to True
-Updated a logging threat to use storesSensitiveData

### DIFF
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -157,7 +157,7 @@ class Element():
         self._name = name
         self._descr = descr
         self._inBoundary = inBoundary
-        self._onAWS = False
+        self._inAWS = False
         self._isHardened = False
         self._inScope = True
         TM._BagOfElements.append(self)
@@ -262,7 +262,7 @@ class Datastore(Element):
         self._onRDS = False
         self._storesLogData = False
         self._storesPII = False
-        self._storesSensitiveData = False
+        self._storesSensitiveData = True
         self._isEncrypted = False
         self._isSQL = True
         self._providesConfidentiality = False
@@ -280,7 +280,17 @@ class Datastore(Element):
         print("{0} [\n\tshape = none;\n\tcolor = {1};".format(_uniq_name(self.name), color))
         print('\tlabel = <<table sides="TB" cellborder="0" cellpadding="2"><tr><td><font color="{1}"><b>{0}</b></font></td></tr></table>>;'.format(self.name, color))
         print("]")
-    
+
+    @property
+    def storesSensitiveData(self):
+        return self._storesSensitiveData
+
+    @storesSensitiveData.setter
+    def storesSensitiveData(self, val):
+        if val not in (True, False):
+            raise ValueError("storesSensitiveData can only be True or False on {}".format(self._name))
+        self._storesSensitiveData = val
+
     @property
     def storesPII(self):
         return self._storesPII
@@ -310,6 +320,17 @@ class Datastore(Element):
         if val not in (True, False):
             raise ValueError("onRDS can only be True or False on {}".format(self._name))
         self._onRDS = val
+
+    @property
+    def isSQL(self):
+        return self._isSQL
+
+    @isSQL.setter
+    def isSQL(self, val):
+        if val not in (True, False):
+            raise ValueError("isSQL can only be True or False on {}".format(self._name))
+        self._isSQL = val
+        self._providesConfidentiality = True
 
     @property
     def isEncrypted(self):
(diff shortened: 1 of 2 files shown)


izar:
Won't storesSensitiveData default to True lead to too many false positives? I think it may be preferable to leave it False and cause people to actually think about data sensitivity when turning it on.

nozmore:
Personally I'd prefer to go with the default of storing sensitive data then have the team think about data sensitivity and set it to False once they have.

nozmore:
I was about to back the storesSensitiveData default value and threat condition change so we can merge the missing getters/setters but I realize that functionally this is no different than the current behavior except that now once someone says storesSensitiveData = False the Threat goes away.

So I think the current code leads to too many false positives, no? 

I am thinking about adding options to configure the default options outside of the pytm.py code, maybe on the TM object itself TM.setDefaultDataStore(DataStore) or maybe some other place where the security team can control it. Then the 'default' here in the code will won't matter since each organization/team could configure it as needed.

izar:
I like the idea. How about making it a command line argument? This way
there'd be flexibility without changing code, and enabling some what-if
scenarios.

nozmore:
Just to be clear, which comment are you referring to? 

As for the storesSensitiveData default value, I think this could merge, there seems to be no real difference as far as increased FPs. 

That said I can see an argument that the alternation to the threat condition is technically a mitigation condition and not a target condition. If the goal is to have all threats to be on the report which apply to a given target then require teams to mitigate them. I actually agree with that goal which is where I was going with the mitigation thread, https://github.com/izar/pytm/issues/18. We could design the model in a way that mitigation have conditions and the report could contain all threats which apply then break them out into Unmitigated, Partially mitigated, Remediated (or sort by severity/cvss, whatever)

Threat
-target list
-target condition (maybe separate target condition for each target type)
-mitigations list
-threat-mitigation condition "(mitigationA or mitigationB) and mitigationC"

Mitigations
-mitigation condition
-severity modification

As for the overall default values paragraph, my thought was to add support for each object type to override the default values in the model to reduce typing and unnecessary complexity in the TM code. 
If I know all my datastores contain sensitiveData, I want all elements inScope=False by default and select the one or two I am modeling, all my datastores are encrypted, or all my elements are in AWS so set inAWS=True. Here is where a report of all elements and annotations come into play so the report can be reviewed to catch things which don't match the default (same is true about the internal defaults too). This is a secondary thing and warrants its own issue to discuss.

izar:
I'm having second thoughts about a wide-ranging default, even if per class. The idea is that the description file, where the model is spelled out, be as descriptive as possible. If we start having settable defaults that may change between models then we start having intrinsic values that are not evident to the eye - i.e., someone who was not the original author that examines the model later on will have to figure out which default setting was used and go from there.
On the other hand if it always starts at a known value and gets changed per element, that's a different story - the starting point is well-known.

nozmore:
I agree on transparency which was why I was thinking of adding the setDefaultDataStore (and similar for each Element) to the tm object so if this is overridden it is visible in the model.

```
defaultDataStore = DataStore()
defaultDataStore.storesSensitiveData = true
tm.setDefaultDataStore(defaultDataStore)
```

These would be visible at the start of the TM model and a report with each element and their state values would provide visibility for verification.

This was more something I was thinking about for future and doesn't need to be ironed out here, let me know your thoughts on the current merge request.

izar (changes requested):
Let's keep storesSensitiveData False right now. I got some people testing this and I don't want to change it under their feet. We can make the default a command-line settable choice.

izar:
Not sure what the semantic value of setDefaultDataStore is - sounds like a data sink? 
I think this complicates things way too much. I don't want to start having implicit "facts" about TMs at this stage.

nozmore:
And backout the change to condition to evaluate the storesSensitiveData?

izar:
Let's think about that for a second - if a datastore stores log data AND it stores sensitive data, are we inferring too much in saying that it is logging sensitive data? I believe we are - it sounds like something that will be a false positive more times than not.

nozmore:
I was looking at this as a log file or device where that wouldn't happen but you could have some FPs if a database has both sensitive and logging data with no overlap between them. 

I'll back out those changes so this can merge.

izar (dismissed):

izar:
Closing since code changed a lot.

PR #11, opened 2018-07-19
added conditional logic for Server/Process shapes based on Scope

nozmore:
I just noticed how the color is set based on inScope so maybe this isn't needed but I use boxes for out of scope components (Server/Process). This code could be moved to a method like _setColor. 

If there are conditions where the original shape is needed we could also set a shape property on the Element so it can be overridden.

I'm submitting this anyway and I can make changes based on PR feedback.

### DIFF
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -174,7 +174,8 @@ class Element():
  
     def dfd(self):
         print("{} [".format(_uniq_name(self._name)))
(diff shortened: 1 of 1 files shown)


izar:
I think I'd like to keep the shape for functionality and the color for scope. It is easier on the head to look at a diagram and "discount" things in a different color than interpret a shape every time anew, plus I think there's value in knowing "what" is the not-in-scope thing.

nozmore:
The difference was so slight I didn't notice it at first, maybe some configuration could be added for the security team to configure the tool to align with their modeling technique.

nozmore:
I have some other ideas for changes which will make everyone happy, closing this...

PR #38, opened 2019-04-03
image location cleanup

davevs:
moved images to seperate folder


izar:
Hi, thanks for the PR, much appreciated!
Couple of comments - "System" might be too higher level an abstraction, no? I mean, Systems are composed of Elements and Servers and ... you get my gist. Perhaps it is the case of re-abstracting it into Element, which was the originally meant base class?

The idea of creating multiple threat files is great! But I think we should aim for a more robust loader mechanism than the eval() call as an argument to .update(). I'm thinking perhaps in the future we could have user-supplied files, and that functionality would benefit from stronger error checking. 

thanks!

davevs:
I agree 'system' is not ideal, but I mostly wanted to clean up the duplicates and moving everything to 'element' would mean 'actors' would also get all the properties (which is not a problem, it just feels weird).

For the threat list loader I'm already testing with JSON files. My idea would be that clients can start from a simple csv file, convert that to JSON (with a tool we provide) and load those into pyTM. I still struggle to make it work with the least amount of changes to the rest of the code though.

izar:
Great - i'll run some testing after locally merging and merge your changes after that. Thanks again!

izar:
This conflicts with some other code we have brewing. I'll close it now and incorporate changes as they fit.

PR #40, opened 2019-06-17
fixed README so that example runs (was missing `Lambda` import)

lojikil:
The _README.md_ file included an example PyTM script that was missing an import; this PR simply adds that import to the example script so that users can easily run the example and see the output from PyTM


izar:
Thank you!

PR #50, opened 2019-10-23
Add cwe/capec references

lfservin:
For most of the threat categories I've tried to link to at least one CWE. Some of them are also attack techniques and thus I've also added in some cases a link to CAPEC. 
I hope it helps make pytm better.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 bin/
 include/
 pip-selfcheck.json
+tm/
 
 # Byte-compiled / optimized / DLL files
 __pycache__/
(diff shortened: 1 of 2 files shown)


izar:
Hi, and thank you so much for this PR!
In the coming days we will be releasing a new version with a new threat format file, and that would overwrite your changes. Would you mind waiting until that is done and then resubmitting again?

izar:
Hi Luis - can you take a look at the changes we have made? We are still working on them - CAPEC/CWE references still have to go into the new format. We should have them by the end of the week. Could you add your additional threats? Thanks!

izar:
If you could resubmit your threats in the new format I'll be glad to merge. Thanks!

PR #54, opened 2019-11-06
Fixing nested Boundaries

jdcallet:


izar:
Thanks for the PR! Could you provide before/after examples of how the issue appears and is resolved by your patch?

pytm/pytm.py
```diff
@@ -524,11 +524,11 @@ def dfd(self):
...
         _debug(result, "Now drawing boundary " + self.name)
+        if type(self) == Boundary:
+            if not self._is_drawn:
+               _debug(result, "Now drawing boundary " + self.name)
+               self.dfd()
         for e in TM._BagOfElements:
-            if type(e) == Boundary:
-                if not e._is_drawn:
-                   _debug(result, "Now drawing boundary " + e.name)
-                   e.dfd()
```
redshiftzero, line 531: good catch - your approach to resolve makes sense. However I think this change can lead to elements being drawn twice due to the fact that `TM._BagOfElements` can contain `Boundary` objects. That is, due to the recursive calling of `Boundary.dfd` in the nested boundary case, we'll end up drawing any elements that are present inside nested boundaries multiple times - each element being drawn once for each boundary it's inside.

To handle this, we could do something like setting an attribute `_is_drawn = False`  on the `Element` object also and set it to true when `Element.dfd()` is called (and we'll need to do the same for any objects that inherit from `Element` and override `Element.dfd()`). Then, when we iterate through `TM._BagOfElements` in the for loop, we check `if not e._is_drawn` and only call `e.dfd` if we've not drawn the element already. What do you think?

redshiftzero: @jdcallet I explored this approach in e65af2027395065b0c6524765293d1bd9149ec9e

izar (approved):

PR #56, opened 2019-11-08
render elements with duplicate names in the DFD

### LINKED ISSUE #45: Boundaries with same name are grouping together
I'm not sure if it's by intent or not, but if I have two boundaries with the same name the components rendered into one boundary.
i.e:
```
dbVpc = Boundary("VPC")
serviceVpc = Boundary("VPC")

...

db = Datastore("Postgres Aurora")
db.inBoundary = dbVpc
server = Server("Server")
server.inBoundary = serviceVpc
```
Generates the following:

image

redshiftzero:
Fixes #45

See reproducer [here](https://github.com/izar/pytm/issues/45#issuecomment-551381567), which now renders objects with duplicate names:

dfd

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,6 +32,8 @@ Sometimes you will find the root cause yourself in the process.
 
 Try to give your issue a title that is succinct and specific. The devs will rename issues as needed to keep track of them.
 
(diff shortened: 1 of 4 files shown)


izar:
Thanks again!

PR #60, opened 2019-11-25
bugfix: 'TypeError: _uniq_name() missing 1 required positional argume…

thneumann:
./tm.py --seq
@startuml
actor cbdcbcbdaaddaadbdfadadfaaae as "User"
File "./tm.py", line 64, in
tm.process()
File "/opt/pytm/pytm/pytm.py", line 255, in process
self.seq()
File "/opt/pytm/pytm/pytm.py", line 234, in seq
print("entity {0} as "{1}"".format(_uniq_name(e.name), e.name))
TypeError: _uniq_name() missing 1 required positional argument: 'obj_uuid'

uuid is missing. 
This is also why I was not able to create sequence diagrams on Windows environment. Now it is again working on both Linux & Windows.

### DIFF
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -229,13 +229,13 @@ class TM():
             if type(e) is Actor:
                 print("actor {0} as \"{1}\"".format(_uniq_name(e.name, e.uuid), e.name))
             elif type(e) is Datastore:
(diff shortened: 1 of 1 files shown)


lareck (commented):
I confirmed that these changes resolved the issue I was seeing.

izar (approved):

PR #67, opened 2020-02-08
allow setting properties on init

nineinchnick:
For elements/flows that have very few properties:
```python
webserver = Server("web server", inBoundary=internal)
request = Dataflow(user, webserver, "some request", dstPort=443)
```

For elements/flows with some common props:
```python
props = {
    "dstPort": 443,
    "protocol": "HTTPS",
}
login = Dataflow(user, webserver, "first request", **props)
do_stuff = Dataflow(user, webserver, "second request", **props)
```

### DIFF
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -187,7 +187,9 @@ class TM():
     _sf = None
     description = varString("")
 
-    def __init__(self, name):
+    def __init__(self, name, **kwargs):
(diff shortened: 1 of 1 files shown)


izar (approved):
Thanks a lot, these patches add a lot of value. This one specially, I can see it opening the door for JSON-defined objects, taking it even one step further.

PR #68, opened 2020-02-13
Allow loading a custom threats file

nineinchnick:
Allow loading a custom threats file by setting a new `TM.threatsFile` attribute.

Added throwing an exception when an attribute is already set, instead of silently ignoring the new value, which could lead to surprising errors.

Added missing unit tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ Element
 
 ```
 
-For the security practitioner, you may add new threats to the `threatlib/threats.json` file:
+For the security practitioner, you may supply your own threats file by setting `TM.threatsFile`. It should contain entries like:
 
 ```json
 {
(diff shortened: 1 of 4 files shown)


izar (commented):
This is great, thank you so much (incidentally do you have a Twitter handle i can add when I post about the latest additions you made?) !
I am wondering if for completeness the name of the threat file added should be part of the report.

nineinchnick:
My Twitter handle is the same as my username here: https://twitter.com/n1neinchnick

Adding threats filename might be a good idea but on the other hand it does already copy selected threat properties. The default template could include threat IDs. The file itself can change after generating the report so I'm not sure how useful that would be, other than just saying a custom one was used instead of the default.

BTW I got more PRs queued up, just need to get them in one by one to avoid conflicts.

Issue #70, opened 2020-02-17
Issue on sequence diagram


lobuhi:
Hi all,

I'm just starting with pytm, following examples and if I try to follow the README instructions I got this using --seq

```
kali@kali:~/pytm$ ./tm.py --seq
@startuml
entity cbfcfffaaffdacebab as "Internet"
entity aeebeaeccccdadbfbbed as "Server/DB"
entity acdbbafaebffabebcefad as "AWS VPC"
actor bdafaaafeadfceac as "User"
database afeaeabaaabeaaabcbedeaa as "SQL Database"
bdafaaafeadfceac -> facdfcdecbdebecaeaa: User enters comments (*)
note left
This is a simple web app
that stores and retrieves user comments.
end note
facdfcdecbdebecaeaa -> afeaeabaaabeaaabcbedeaa: Insert query with comments
note left
Web server inserts user comments
into it's SQL query and stores them in the DB.
end note
afeaeabaaabeaaabcbedeaa -> facdfcdecbdebecaeaa: Retrieve comments
facdfcdecbdebecaeaa -> bdafaaafeadfceac: Show comments (*)
bbeedeaacfffcbbbbcba -> afeaeabaaabeaaabcbedeaa: Lambda periodically cleans DB
@enduml

kali@kali:~/pytm$ ./tm.py --seq | java -Djava.awt.headless=true -jar plantuml.jar -tpng -pipe > seq.png
Picked up _JAVA_OPTIONS: -Dawt.useSystemAAFontSettings=on -Dswing.aatext=true
```

And then this is my seq diagram
![image](https://user-images.githubusercontent.com/37011545/74645033-8571b080-5177-11ea-848a-600ad286e26a.png)

Am I missing something? I've already check that I've all requirements installed.

Thx in advance ^_^

luixtao:
Ok, got it! 

Why Server or Lambda instances are not defined in sequence instruction?

 ```
   def seq(self):
        print("@startuml")
        for e in TM._BagOfElements:
            if isinstance(e, Actor):
                print("actor {0} as \"{1}\"".format(_uniq_name(e.name, e.uuid), e.name))
            elif isinstance(e, Datastore):
                print("database {0} as \"{1}\"".format(_uniq_name(e.name, e.uuid), e.name))
            elif not isinstance(e, Dataflow) and isinstance(e, Boundary):
                print("entity {0} as \"{1}\"".format(_uniq_name(e.name, e.uuid), e.name))
```

[...]

PR #71, opened 2020-02-17
Avoid boundary icons on sequence diagram as an entity

lobuhi:
Modified conditional in line 221 to avoid boundary icon drawn as an entity on sequence diagram.
Check [#70](https://github.com/izar/pytm/issues/70)


izar:
Thanks!

PR #73, opened 2020-02-20
adding json output in pytm

cmorell:
Hi,

I really liked your project but for me would be helpful to has a json output. I have done it, take a look 

Thanks!


izar:
Hi, thanks for the patch!
How is this different from writing a report template that outputs JSON ?

{
  'name':'{tm.name}'',
  'description':'{tm.description}'',
  'threats':[
    {findings:repeat:
      'finding': '{item.id}',
      'description': '{item.description}',
      .....
    }
  ]
}

cmorell:
You're right. I change the way to generate report using template.json file in docs/template.json 
Now I think it's better (no new argument needed)

izar (commented):
Can you provide  a PR straight from master? I'd prefer not to commit code just to remove it.

PR #74, opened 2020-02-21
add json template

cmorell:
Adding json template
now, PR straight to master


izar:
It would be really preferable not to introduce format-dependent code into the template engine. We are aiming to have it as simple as possible. Can't JSON be correctly expressed by using just the template?

cmorell:
Hi, 

I understand your point, but I found problems with right/left curly brace with the current template_engine. 
Also with the last comma to separate all the potential threats inside the list of findings

cmorell:
I mean, I've tested to add html entities (&lcub;  or &lbrace;) and is didin't work. If I use "{" I think template engine interpret it like a start of a function, like "{findings:repeat:"

izar:
Perhaps we can fix that then and add escapable curly braces. I would really
prefer to not introduce format-based logic.

izar:
I just took a look at it. Since the template engine is based on format(),
you can just double the curly braces to get it working. For example:

{{{tm.name}}}

prints out:

{my test tm}

if run as  template with the sample tm.py. Can you try your JSON template
with this in mind?

thanks!

cmorell:
Hi, 
I've found it and more or less solve the problem, but, the problem now is the comma to separate elements inside the array of findings.

Solution found in: https://stackoverflow.com/questions/51562600/escaping-curly-braces-in-string-to-be-formatted-an-undefined-number-of-times

```
{{
"name":"{tm.name}",
"description":"{tm.description}",
"potential threats":
[
{findings:repeat:
⁍
	"name":   "{{item.id}}",
	"description":"{{item.description}}",
	"targeted element": "{{item.target}}",
	"severity" : "{{item.severity}}",
	"example instances":"{{item.example}}",
	"mitigations":"{{item.mitigations}}",
	"references":"{{item.references}}"
⁌, 
}
]
}}
```

After it, we need to change last occurrence of }, to } to make a valid json

nineinchnick:
If you need a JSON output you should generate it in Python after building the model, not using a template:
```python
tm = TM("some model")
# add assets and dataflows
tm.resolve()

data = {
  "name":tm.name,
  "description":tm.description,
  "threats": [{"name": f.id} for f in TM._BagOfFindings],
}

with open('report.json', 'w') as f:
    json.dump(data, f)
```

cmorell:
Hi @nineinchnick 
It was the first approach, but @izar suggest me to use template option.
For me go back and use your approach will be easy

Just let me know

nineinchnick:
Generating structured data using an unstructured template creates a risk of the output file to be invalid, because it's easy to make a typo in the template or not properly escape template data. If you save a JSON file from Python it's always going to be a correct JSON file and whatever you feed it to will be able to parse it. If you generate it using a template and make a typo, like a missing brace or extra comma, it's hard to pin point exactly where is it. That's the issue you're currently having - an extra comma.

nineinchnick:
#102 should also address this but in a more robust way

PR #76, opened 2020-02-26
merge graph edges

### LINKED ISSUE #53: bidirectional dataflows
when an element can be both a source and a sink with respect to a given data flow - for example, a user interacting with a web application where they both can download and upload data - it would be nice to define a single `Dataflow` object (or `BidirectionalDataflow` if another class is preferred) that renders as a bidirectional dataflow in the DFD and elicits the threats in both directions. is there interest in supporting that?

nineinchnick:
Allow to collapse a request and response dataflow into a single double-arrow edge. This greatly simplifies bigger DFDs and makes them more (actually) readable.

Added 3 new Dataflow attributes:
* isResponse - bool, when set and there's exactly one other dataflow for the same source and sink in reversed direction treat it as the request
* responseTo - point to another dataflow that this one is a response to
* response - point to another dataflow that's a response to this one

If only one of those attributes is set, others are filled in automatically, when possible.

When `TM.mergeResponses` is set, the resulting DFD would look like:
dfd

Possibly closes #53

### DIFF
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -1,6 +1,7 @@
 import argparse
 import json
 import uuid
+from collections import defaultdict
 from hashlib import sha224
 from os.path import dirname
 from re import match, sub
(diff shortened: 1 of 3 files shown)


izar:
Wow, I don't know. To me it makes it harder to read. Let me ask the rest of the team what they think.

nineinchnick:
The example might be too simple. The point is that this enables creating multiple models reusing some, but not all assets. It also allows defining multiple models in the same package, for example if you'd want to add some advanced validation logic that requires to verify multiple models at once. The model I'm working on is so big I need to split it up into multiple smaller ones.

nineinchnick:
The example might be too simple. It works better for large models, especially when there are multiple data flows between one pair of assets. Also note this is optional and not enabled by default.

PR #81, opened 2020-03-05
allow advanced templates in repeat blocks

nineinchnick:
Allows to call methods on items in repeat blocks (notice `item.target.display_name:call`):
```
{findings:repeat:|{{item.target.display_name:call}}|{{item.id}}|{{item.description}}|{{item.severity}}|{{item.details}}|{{item.mitigations}}|{{item.example}}|{{item.references}}
}
```
We could use the property decorator but there's no reason why the slightly more advanced template format shouldn't be supported in the repeat block.


izar:
Thanks!

PR #86, opened 2020-03-15
assign findings to elements

nineinchnick:
Assign findings to elements to allow to render them in reports grouped by assets/dataflows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,29 @@ Name|From|To |Data|Protocol|Port
 
 ```
 
+To group findings by elements, use a more advanced, nested loop:
+
+```text
+## Findings
+
+{elements:repeat:{{item.findings:if:
+### {{item.name}}
+
+{{item.findings:repeat:
+**Threat**: {{{{item.id}}}} - {{{{item.description}}}}
+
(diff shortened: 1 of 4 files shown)


nineinchnick:
Here's an example template. Pretty ugly but works:
```
## Findings

{elements:repeat:{{item.findings:if:
### {{item.name}}

{{item.findings:repeat:
**Threat**: {{{{item.id}}}} - {{{{item.description}}}}

**Severity**: {{{{item.severity}}}}

**Types**: {{{{item.categories}}}}

**References**: {{{{item.references}}}}

}}}}}

```

Finding properties need to be double escaped since there are two loops, hence four brackets. Fortunately the `if` doesn't add another level of escaping. There are five closing brackets at the end since two close the inner loop, next two the `if` and last one closes outer loop.

I think this is as far as this templating system should be taken but grouping by asset/dataflow should make it easier to manage large number of findings.

izar (commented):
this one looks good - can you add a sample template ?

nineinchnick:
Done!

izar (approved):

PR #87, opened 2020-03-15
support running in a docker container

nineinchnick:
New make targets and vars:
* `make image` builds a `pytm` Docker image
* `make tm/report.html` or any other file (dfd.png, seq.png) to build single files, useful if having multiple models but working only on one
* `make report USE_DOCKER=true` runs everything in a Docker container

Targets correctly recognize file timestamps so only results for updated files are being built.

> NOTE: I can't test this on OSX as I don't have a mac

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -120,3 +120,5 @@ tm_example.dot
 .idea
 *.iml
 
+plantuml.jar
+tm/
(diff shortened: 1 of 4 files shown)


izar (approved):
Thanks!

PR #88, opened 2020-03-17
move authenticateDestination to base Element

nineinchnick:
Move authenticateDestination to base Element and make Dataflows inherit it from sources to properly model when clients do/don't verify server certificates.


izar (approved):
Need to add this to CHANGELOG as I think it breaks existing scripts silently.

PR #89, opened 2020-03-20
assign inputs and outputs to all elements

nineinchnick:
Assign inputs and outputs to all elements. This allows creating threats with more fine grained conditions. A most basic condition is to match only elements with incoming dataflows (open ports) using `any(target.inputs)`. A more advanced is to match clients of particular services by inspecting all outgoing dataflows: `any(f.sink.oneOf(Datastore) and f.sink.isSQL for f in target.outputs)`


izar (approved):
Good to go, thanks!

PR #93, opened 2020-03-26
ensure all items has correct color (based on scope)

nineinchnick:
Ensure all items has correct color (based on scope). Currently actors not in scope are drawn as black, which is confusing considering other items out of scope are gray.


izar (approved):
For some reason fontcolor didn't work for me way back then. Hopefully this is a good change on graphviz's part.

PR #94, opened 2020-03-26
use numbered dataflow labels on seq diagram

nineinchnick:
Use numbered dataflow labels on seq diagram. This makes it easier to read when comparing seq and dfd diagrams.

Please note that changes here are very similar to what's proposed in #80, that is:
* moving label and color functions to methods of the `Element` class
* adding shape and display_name methods
which makes it even easier to override classes. As a side effect, `_dfd()` method has been removed from most classes since it's a duplicate now.

This also has the same fix as #93 but that other PR is way simpler and might be merged earlier.

Preview:
seq

### DIFF
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -18,7 +18,9 @@ from .template_engine import SuperFormatter
 
 ''' Helper functions '''
 
-''' The base for this (descriptors instead of properties) has been shamelessly lifted from https://nbviewer.jupyter.org/urls/gist.github.com/ChrisBeaumont/5758381/raw/descriptor_writeup.ipynb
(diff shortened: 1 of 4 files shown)


pytm/pytm.py
```diff
@@ -425,18 +417,28 @@ def _uniq_name(self):
...
+        print('\tlabel = <<table border="0" cellborder="0" cellpadding="2"><tr><td><b>{0}</b></td></tr></table>>;'.format(self._label()))
         print("]")
 
+    def _color(self):
+        if self.inScope is True:
+            return "black"
+        else:
+            return "grey69"
+
+    def display_name(self):
```
nineinchnick: We can consider defining those with the `@property` decorator.

PR #96, opened 2020-04-02
document attributes that are acronims

nineinchnick:
I had to look up what's POLP so I copied a few definitions from Wikipedia.

### DIFF
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -458,7 +458,12 @@ class Element():
     onAWS = varBool(False)
     isHardened = varBool(False)
     implementsAuthenticationScheme = varBool(False)
-    implementsNonce = varBool(False)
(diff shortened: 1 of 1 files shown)


pytm/pytm.py
```diff
@@ -670,7 +682,12 @@ class Datastore(Element):
     authenticationScheme = varString("")
     usesEncryptionAlgorithm = varString("")
     validatesInput = varBool(False)
-    implementsPOLP = varBool(False)
+    implementsPOLP = varBool(False, doc="""The principle of least privilege (PoLP),
```
nineinchnick, line 685: To avoid duplicating these doc strings, we might consider adding a base class for Datastore, Process and Server since they share so many attributes but we might now want to all of them to the Element. Another approach would be to keep them in a global var, but I'd prefer proper inheritance than polluting the NS with global vars.

izar (approved):

PR #98, opened 2020-04-10
Addition of JSON as new input type

avhadpooja:


izar (changes requested):
I am still having difficulties with understanding the use case for this one. Who wakes up in the morning thinking "oh i think i'll write myself a system description in JSON, it is so much clearer than using statements!" ? 
How about making this a utility script that transforms well-formed JSON into python code that runs pytm? This way the code stays focused and anyone wanting to move from JSON to code is able to have the middle format to work from.

nineinchnick:
Possibly superseded by #105

PR #99, opened 2020-05-19
handle duplicate dataflows

nineinchnick:
Some models might have multiple Dataflows between same source and sink that have all same properties (except name and note). They may be required if using the sequence diagram to make it complete but drawing them on the DFD makes it less readable.

This MR adds a new property called `TM.onDuplicates` with three possible values. By default, no action is taken. When set to `RESTRICT`, an exception is raised when such dataflows are found. This is meant for models not using the sequence diagram. When set to `IGNORE`, duplicates are allowed but are not drawn on DFD, only on the sequence diagram.


izar (approved):

izar:
Thanks!

PR #105, opened 2020-09-11
load model from a json file

nineinchnick:
since #102 adds json output, this PR adds json input, for example if models can be created by exporting data from other systems


izar (commented):
Can you take a look at the conflicts? I am not sure if you renamed or added to the _Bag* variables.

nineinchnick:
I rebased both this and #102 which is based on this branch and should be merged first, it'll make this diff a lot smaller.

PR #106, opened 2020-09-22
Fixed setup.py to install deps

archen:
Added the install_requires line needed to install pypi dependencies and
opportunistically fixed the usage of `summary` which isn't recognized by
dist/setuptools and used `description` instead.


nineinchnick (approved):

nineinchnick (approved):

izar:
Thanks! Interesting though, the previous version was passing Github CI checks.

PR #116, opened 2020-09-29
Issue #115: provideIntegrity missing on Actor and Asset class

### LINKED ISSUE #115: Crash on checking threat attributes that are not in current object while generating report
### How does the problem look like and what steps reproduce it?
Issue can be easily reproduced when trying to generate report using provided threat library, sample tm.py (both one in repo and another slightly different in README.md) and template.
Traceback using tm.py from repo:
```
Exception has occurred: AttributeError
'Actor' object has no attribute 'providesIntegrity'
  File "/root/pytm/pytm/pytm.py", line 445, in apply
    return eval(self.condition)
  File "/root/pytm/pytm/pytm.py", line 547, in resolve
    if not t.apply(e):
  File "/root/pytm/pytm/pytm.py", line 721, in process
    self.resolve()
  File "/root/pytm/tm.py", line 91, in <module>
    tm.process()
```
Threat being checked is `AC05` with condition `'((not target.source.providesIntegrity or not target.sink.providesIntegrity) and not target.isEncrypted) or (target.source.inScope and not target.isResponse and (not target.authenticatesDestination or not target.checksDestinationRevocation))'`. As we know `Actor` object doesn't have any `providesIntegrity` attribute, but it's being checked.
### Can you reproduce it using the latest master?
Yes. That's what I used.
### What is your running environment? 
OS: SLES 15/python:alpine-3.8 image 
Python version: 3.6.10/3.8.6
Your model file, if possible: sample `tm.py` from repo and another one from `README.md`
### What have you already tried to solve the problem?
Not yet. I'm not proficient in Python and still poking the code.
EDIT: I think a simple exception can be added to handle such attribute issues in non-elegant way:
```python
    def apply(self, target):
        if not isinstance(target, self.target):
            return None
        try:
            return eval(self.condition)
        except AttributeError:
            return None
```

izar:


pytm/pytm.py
```diff
@@ -1144,6 +1145,8 @@ class Actor(Element):
     checksDestinationRevocation = varBool(False, doc="""Correctly checks the revocation status
 of credentials used to authenticate the destination""")
     isAdmin = varBool(False)
+    # should not be settable, but accessible
```
nineinchnick, line 1148: Why it should not be settable? An actor can sign requests

nineinchnick: How about moving `providesIntegrity` to the base class?

izar: an actor signs a request via a process on their behalf, which should be looked at as a Process. I want to keep the Actor semantically separated (right now it is not) from the process it uses in order to allow us to use it for privacy threats later on.

nineinchnick: Using a plain bool doesn't solve this anyway. I think we should add `not target.source.oneOf(Actor) or not target.source.providesIntegrity` (and same for sink) in the thread condition.

izar: we probably need to implement a "unless_target()" filter or something for these cases.
you mean the plain bool doesn't solve the existing-but-not-settable part, or is there something beyond that I am not seeing here?

izar: TODO: revisit Element/Asset and how the attributes are inherited, clean up inheritance and come up with a cleaner, better method to treat these going forward.

PR #119, opened 2020-10-02
delete unused attributes

nineinchnick:
The following attributes are not used in any threat. Using them in models can create a false sense of security where no new findings would be reported.

- authenticatedWith
- authenticationScheme
- codeType
- handlesCrashes
- handlesInterruptions
- hasWriteAccess
- implementsCommunicationProtocol
- isAdmin
- isSQL
- onAWS
- onRDS
- OS
- providesConfidentiality
- storesLogData
- storesPII
- storesSensitiveData
- tracksExecutionFlow

Its pretty easy for anyone to extend pytm classes and add these back when referencing them in custom threats. I believe that extending pytm this way should be encouraged.

This is controversial but it's easier to remove unused attributes than try to figure out correct description for them. Less is more.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,6 @@ The available properties of an element can be listed by using `--describe` follo
 
 (pytm) ➜  pytm git:(master) ✗ ./tm.py --describe Element
 Element class attributes:
-  OS
   definesConnectionTimeout        default: False
   description
   handlesResources                default: False
@@ -53,10 +52,8 @@ Element class attributes:
   implementsNonce                 default: False
   inBoundary
   inScope                         Is the element in scope of the threat model, default: True
-  isAdmin                         default: False
   isHardened                      default: False
   name                            required
-  onAWS                           default: False
 
 ```
 
@@ -83,14 +80,11 @@ user = Actor("User")
 user.inBoundary = User_Web
 
 web = Server("Web Server")
-web.OS = "CloudOS"
 web.isHardened = True
 
 db = Datastore("SQL Database (*)")
-db.OS = "CentOS"
 db.isHardened = False
 db.inBoundary = Web_DB
-db.isSql = True
 db.inScope = False
 
 my_lambda = Lambda("cleanDBevery6hours")
@@ -261,7 +255,7 @@ If `target` is a Dataflow, remember you can access `target.source` and/or `targe
 Conditions on assets can analyze all incoming and outgoing Dataflows by inspecting
 the `target.input` and `target.output` attributes. For example, to match a threat only against
 servers with incoming traffic, use `any(target.inputs)`. A more advanced example,
-matching elements connecting to SQL datastores, would be `any(f.sink.oneOf(Datastore) and f.sink.isSQL for f in target.outputs)`.
+matching elements connecting to SQL datastores, would be `any(f.sink.oneOf(Datastore) for f in target.outputs) and target.protocol == 'SQL'`.
 
 ## Currently supported threats
 
--- a/docs/pytm/index.html
+++ b/docs/pytm/index.html
@@ -198,7 +198,6 @@ for example by verifying the authenticity of a digital certificate.&#34;&#34;&#3
         doc=&#34;&#34;&#34;Correctly checks the revocation status
 of credentials used to authenticate the destination&#34;&#34;&#34;,
     )
-    isAdmin = varBool(False)
     # should not be settable, but accessible
     providesIntegrity = False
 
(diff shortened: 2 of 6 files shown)


nineinchnick:
These should be moved to the new Data class (without the stores prefix) with another new attr like `isStored`, to separate dataflows for select and insert/update queries.
- storesLogData
- storesPII
- storesSensitiveData

izar:
You're forgetting the dimension that an attribute, in and off itself, can serve as a documentation/clarification. Granted, they are not part of a threat right now but they can offer clarity. We need a better job at documentation and using enumerations for possible values where appropriate, but i think these attributes still have a place.
About the stores* I agree that they have a place in Data, but i would say that it is equally important to be able to pinpoint when a store stores a given type of data - either as clarification of for rules that are more datastore oriented than data itself.

nineinchnick:
> You're forgetting the dimension that an attribute, in and off itself, can serve as a documentation/clarification. Granted, they are not part of a threat right now but they can offer clarity. We need a better job at documentation and using enumerations for possible values where appropriate, but i think these attributes still have a place.

I hold my argument that if these attrs are not documented and not used in any threats they only add confusion when starting out with `pytm` and building first models. The names are not obvious enough and users can waste lots of time trying to figure out the "correct" value only to later learn that this was not necessary.

You can still add custom attributes to any object and reference them in the template. Since you'd have to reference them in your custom template I don't see any value in providing any default "doc only" attributes.

I only see two options, either removing them or documenting them. If you want to do the latter, please open up a PR and I would remove documented attrs from this one.

> About the stores* I agree that they have a place in Data, but i would say that it is equally important to be able to pinpoint when a store stores a given type of data - either as clarification of for rules that are more datastore oriented than data itself.

Marking a datastore as storing a particular kind of data only makes sense if there are any incoming dataflows that carry this kind of data. Even if we'd keep these `stores*` attributes, they should be populated automatically. I can add that in this PR if you agree.

izar:
> I hold my argument that if these attrs are not documented and not used in any threats they only add confusion when starting out with pytm and building first models. The names are not obvious enough and users can waste lots of time trying to figure out the "correct" value only to later learn that this was not necessary.

Totally agreed. We MUST have better docs, and we are ok to change names.

> Marking a datastore as storing a particular kind of data only makes sense if there are any incoming dataflows that carry this kind of data.

Consider a project where many developers are creating the model. That exact discrepancy is something worth flagging out. We cannot assume that the same person creating the datastore is the one documenting the dataflows.

colesmj:
There is a value to allowing users to set custom attributes, or to set
attributes that make sense for an object or flow but that does not _yet_
contribute to a threat. But perhaps this something we solve in v2, not
currently.


~Matthew Coles~

nineinchnick:
@izar @colesmj so what's the conclusion? Should we remove attrs that are not well defined or try to document them? Removing is easier, hence this PR.

izar:
I'm for documenting rather than deleting.

nineinchnick:
> I'm for documenting rather than deleting.

I can give a shot for the obvious ones like OS but:
1. what about attrs that are completely ambiguous, like `handlesCrashes`? How does it handle crashes? What kind of crashes?
1. how to document that some attrs are not used in any threats? In their docs or should we keep a separate list somewhere?
1. should docs include example values? Like for OS should it be Linux, Windows, OSX or more specifically distros and versions?

izar:
Of course some can be deleted if we decide they're not useful :)
For example, for handlesCrashes, i could imagine something like a named bit
map (
https://docs.astropy.org/en/stable/api/astropy.nddata.bitmask.BitFlagNameMap.html#astropy.nddata.bitmask.BitFlagNameMap)
with OS, Process, etc. to express at which level(s) the crash may be
handled. That would also cover example values. Enumerations and bit masks
would make things clearer IMHO.

nineinchnick:
> 
> Of course some can be deleted if we decide they're not useful :) For example, for handlesCrashes, i could imagine something like a named bit map ( https://docs.astropy.org/en/stable/api/astropy.nddata.bitmask.BitFlagNameMap.html#astropy.nddata.bitmask.BitFlagNameMap) with OS, Process, etc. to express at which level(s) the crash may be handled. That would also cover example values. Enumerations and bit masks would make things clearer IMHO.

Sorry but that's still very vague. `pytm` should not try to model every property of every possible system since this is impossible. We should only keep things that are in any way related to threats. That's why I suggest removing all the attributes mentioned in this PR. Including generic things like OS if we don't have OS specific threats.

izar:
I do believe being able to say that a specific element handles crashes both
at the OS level (with some kind of reboot watchdog for example) or at the
application level (by running in a loop, for example) is important
information to have.
And yes, things will be vague until they're solidified either by
documentation or added to a threat or... - it is the nature of the beast.

nineinchnick:
> I do believe being able to say that a specific element handles crashes both at the OS level (with some kind of reboot watchdog for example) or at the application level (by running in a loop, for example) is important information to have. And yes, things will be vague until they're solidified either by documentation or added to a threat or... - it is the nature of the beast.

I've chosen a bad example since `handlesCrashes` just begs to have a threat added and I'm pretty sure we'll find some matching CWEs quickly. So I can sign up for handling this one just to keep things moving forward. But what about others? Can we go through all of them one by one? Should we take this to Slack?

izar:
Sure thing, let's see what makes sense to keep and what to drop. We might
even draw an informal policy on what and how to add going forward.

colesmj:
I will say that we should be maintaining a data model. I started one in the
wiki in GitHub we can leverage (it needs updating of course for all the
work that has been done so far).
Let me know if you want me to take this as an action.

~Matthew Coles~

colesmj:
FYI. I went ahead and I've started on the object model for 1.x release in
the wiki based on the code and documentation within.

~Matthew Coles~

izar:
That's a great idea. We should use that as the base for documentation - the
"source of truth" going forward.

PR #120, opened 2020-10-02
install requirements in docker image

nineinchnick:
install requirements in docker image

the whole hacktoberfest thing makes me feel slightly guilty for doing small PRs like this :-P


izar:
Small is manageable :)

PR #122, opened 2020-10-02
format all files with black

nineinchnick:
format all files with black


izar (approved):
It looks like black likes empty lines in funny places.

PR #123, opened 2020-10-02
check for min tls version

nineinchnick:
Check for min tls version, similar to how we check for max classification.

Remove `usesLatestTLSversion` since it's not precise enough what's "latest".


pytm/pytm.py
```diff
@@ -830,6 +848,12 @@ class Element:
         required=False,
         doc="Maximum data classification this element can handle.",
     )
+    minTLSVersion = varTLSVersion(
+        TLSVersion.NONE,
+        required=False,
+        doc="""Minimum required TLS version required.
+Note that currently only TLS 1.2 and 1.3 are considered secure.""",
```
izar: just wondering if the comment about the versions is not better suited for the actual threat. First, the versions will probably move and we will have to chase them here, second, on a more meta level, this adds a level of judgement that i don't think is present in other attributes (off memory, I didn't check)

izar: Yep, I think this needs changing.

izar (commented):
Just need to fix that message.

pytm/pytm.py
nineinchnick: I removed the note and I noticed that `required` is used twice.

PR #127, opened 2020-10-04
mark data as pii or creds and check encryption at rest

nineinchnick:
* mark data as pii or creds and if it's encrypted at rest
* add two new threats, one that checks if sensitive data is sufficiently protected, second to check if credentials are not hardcoded (or otherwise long lived)

This is supposed to allow removing attrs like `storesPII` from the Datastore in #119


rgacote:
If PII is explicitly identified, why not also break out PCI (credit cards) and PHI (health information)?

nineinchnick:
> If PII is explicitly identified, why not also break out PCI (credit cards) and PHI (health information)?

@rgacote This is a great idea. Would you like to work on it? At minimum this involves:
* defining new attributes and documenting them
* using them in conditions of existing threats or defining a new threat

rgacote:
> > If PII is explicitly identified, why not also break out PCI (credit cards) and PHI (health information)?
> 
> @rgacote This is a great idea. Would you like to work on it? At minimum this involves:
> 
> * defining new attributes and documenting them
> * using them in conditions of existing threats or defining a new threat

Wish I could work on it. In the middle of three security assessments at the moment...

PR #132, opened 2020-10-16
Added levels to DFDs

izar:
This is and idea to be able to add to the same model, for example, infra-structure and application layers, then select on which diagram to produce. This only impacts DFDs, threat elicitation, sequence diagrams and reporting stay the same.

### DIFF
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -94,6 +94,18 @@ class varInt(var):
         super().__set__(instance, value)
 
 
+class varInts(var):
+    def __set__(self, instance, value):
+        if not isinstance(value, Iterable):
+            value = [value]
+        for i, e in enumerate(value):
+            if not isinstance(e, int):
+                raise ValueError(
+                    f"expecting a list of int, item number {i} is a {type(e)}"
+                )
+        super().__set__(instance, set(value))
+
+
 class varElement(var):
     def __set__(self, instance, value):
         if not isinstance(value, Element):
@@ -156,20 +168,6 @@ class varData(var):
                     )
                 )
         super().__set__(instance, DataSet(value))
-
-
-class varListOfInts(var):
-    def __set__(self, instance, value):
-        if isinstance(value, int):
-            value = [int(value)]
-        if not isinstance(value, Iterable):
-            value = [value]
-        for i, e in enumerate(value):
-            if not isinstance(e, int):
-                raise ValueError(
-                    f"expecting a list of int, item number {i} is a {type(e)}"
-                )
-        super().__set__(instance, set(value))
 
 
 class DataSet(set):
@@ -323,6 +321,8 @@ def _apply_defaults(flows, data):
             processors[d].add(e.source)
             processors[d].add(e.sink)
 
+        e._safeset("levels", e.source.levels & e.sink.levels)
+
         if e.isResponse:
             e._safeset("protocol", e.source.protocol)
             e._safeset("srcPort", e.source.port)
@@ -416,19 +416,6 @@ def _get_elements_and_boundaries(flows):
                 elements.add(b)
                 boundaries.add(b)
     return (list(elements), list(boundaries))
-
-
-def _render_by_level(level):
-    # "only Level 0" (default setting) always renders everything
-    if TM._levels == [0]:
-        return True
-
-    # if the user requests other levels than just 0, deal with it
-    for l in level:
-        # if one of the levels of the Element appears in the user-requested level list, it renders
-        if l in TM._levels:
-            return True
-    return False
 
 
 """ End of help functions """
@@ -533,7 +520,6 @@ class TM:
     """Describes the threat model administratively,
     and holds all details during a run"""
 
-    _levels = [0]
     _flows = []
     _elements = []
     _threats = []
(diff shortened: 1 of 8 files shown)


pytm/pytm.py
```diff
@@ -404,6 +404,18 @@ def _get_elements_and_boundaries(flows):
...
 
 
+def _render_by_level(level):
+    # Level 0 (default) always renders
+    if TM._levels == [0]:
+        return True
+
+    for l in level:
+        # if one of the levels of the Element appears in the user-requested level list, it renders
+        if l in TM._levels:
```
nineinchnick: This should be a method of `TM` and should refer to `self`. Let's stop using global variables

pytm/pytm.py
```diff
@@ -404,6 +404,18 @@ def _get_elements_and_boundaries(flows):
     return (list(elements), list(boundaries))
 
 
+def _render_by_level(level):
+    # Level 0 (default) always renders
+    if TM._levels == [0]:
```
nineinchnick: The default of `_levels` could be an empty list meaning all levels.

pytm/pytm.py
```diff
@@ -676,6 +689,11 @@ def dfd(self):
                 if e.response is not None:
                     e.response._is_drawn = True
         for e in TM._elements:
+            sys.stderr.write(f"{e.name}\t\t\t\t")
```
nineinchnick: Use the logger, there are example calls to it in other places here.

pytm/pytm.py
```diff
@@ -1001,6 +1028,12 @@ def __repr__(self):
     def __str__(self):
         return "{0}({1})".format(type(self).__name__, self.name)
 
+    def dfd(self):
```
nineinchnick: Why this is needed? We shouldn't attempt to draw `Data`

pytm/pytm.py
```diff
@@ -1481,6 +1523,12 @@ def get_args():
         "--describe", help="describe the properties available for a given element"
     )
     _parser.add_argument("--json", help="output a JSON file")
+    _parser.add_argument(
```
nineinchnick, line 1527: Huh, github doesn't let me comment on any line in this file. Anyway, on line 1498 we serialize all elements:
```
value = [v.id if isinstance(v, Finding) else v.name for v in value]
```
the `else` block assumes all list items are objects and it fails on levels now. You can check for property name in the parent condition:
```
            if isinstance(value, (Element, Data)):
                value = value.name
            elif isinstance(obj, Threat) and i == "target":
                value = [v.__name__ for v in value]
            elif (
                not nested
                and i != "_levels"
                and not isinstance(value, str)
                and isinstance(value, Iterable)
            ):
                value = [v.id if isinstance(v, Finding) else v.name for v in value]
```

nineinchnick (commented):
I wonder if this should be done by layering multiple models over each other. It would imply that each layer should be a valid model on its own.

Isn't it currently possible to draw the DFD in a way it's missing crucial information to understand the model?

Anyway, this PR is missing tests that would document the use cases for this feature.

pytm/pytm.py
izar: Good stuff, thanks. I'll get into the fixes right away. The sys.stderr was just my silly debugging that i forgot to erase.

The models can be done in overlaying layers, by making the elements participate in different levels at the same time. Yes, we can get a model that misses crucial information on the diagram (the rest of the info is always evaluated) but that's on the owner of the diagram, not us - right now we don't have any form of logical verification of the diagram. That would be a significant step ahead if we could get done, but even then there will be limits.

nineinchnick: Ok I know what feels odd here. Assigning levels manually to every element is very different from how the model is being constructed initially, by defining elements and dataflows with attribute values. Levels should be contexts like in https://github.com/mingrammer/diagrams/blob/master/docs/getting-started/examples.md

pytm/pytm.py
```diff
@@ -838,6 +862,9 @@ class Element:
         doc="Maximum data classification this element can handle.",
     )
     findings = varFindings([])
+    levels = [
```
nineinchnick: Why this is not a `var`?

pytm/pytm.py
izar: that would (in my head) make it difficult to have an element belong to more than one model level. perhaps a 2.0 improvement?

pytm/pytm.py
izar: at some point i had a varListOfInts. Things broke and I couldn't come up with a good rationale for forcing it in.

pytm/pytm.py
izar: Perhaps doesn't belong in this PR. I was thinking that in the future we could have representations of the data, per confidentiality level, in the model.

pytm/pytm.py
izar: I decided to use "0" for all levels due to the nomenclature regarding DFD "zooming" - level 0 is the top level dfd.

nineinchnick: But it's only one level

pytm/pytm.py
nineinchnick: Contexts could be nested. It would also group items in code which would help manage all available levels and ensure elements are not omitted. But I'm fine with doing this in a follow-up PR, I could help with that.

pytm/pytm.py
nineinchnick: An important rationale is to be consistent with other attributes. Levels are not special. You could use `varData` as an example, just with `int` instead of `Data` and a regular set at the end instead of `Dataset`

izar:
Will do!

izar:
Ah. I think i get what you're saying. Let me rethink it.

pytm/pytm.py
izar: Forget about, I confused myself. When TM._levels is just [0] it means there was no user choice via --levels so it is fair to just return true. I've cleared it up in the comments.

pytm/pytm.py
izar: Silly debugging code, removed.

pytm/pytm.py
izar: Open to suggestions here. In pytm.py we don't have an instance of TM to work from, so I was getting a positional argument error when I made it into a method of TM.

nineinchnick: Sorry, I don't get it. If you'll move it to be a method of the TM class you'll have self as the first argument.

pytm/pytm.py
nineinchnick: These are still here, can you remove them?

pytm/pytm.py
izar: I tried. Then once I moved the call to TM._ ... I got the positional argument error. Fishing around showed that it was expecting an instance of TM to do the calling, not TM. The only thing I saw that made it different from the other functions in TM was that it was the one being called from outside.

pytm/pytm.py
```diff
@@ -838,6 +862,7 @@ class Element:
         doc="Maximum data classification this element can handle.",
     )
     findings = varFindings([])
+    levels = varInts({0})
```
nineinchnick, line 865: One last request - this needs a `doc`, can you add it?

pytm/pytm.py
izar: Done.

PR #134, opened 2020-10-30
Adding a configuration file

izar:
Currently the file only supports excluded threats.
Locations: LIB/config.pytm, ~/.config.pytm, ./.config.pytm and ./{tm_file_name}.pytm.
Format:

[Default]
exclude = THREAT1,THREAT2,...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,11 @@ report: $(models:.py=/report.html) seq dfd
 
 .PHONY: test
 test:
+	@mv ~/.config.pytm ~/.config.pytm.test
+	@mv ./.config.pytm ./.config.pytm.test
 	@python3 -m unittest
+	@mv ~/.config.pytm.test ~/.config.pytm
+	@mv ./.config.pytm.test ./.config.pytm
 
 .PHONY: describe
 describe:
--- a/Makefile
+++ b/Makefile
@@ -45,11 +45,7 @@ report: $(models:.py=/report.html) seq dfd
 
 .PHONY: test
 test:
-	@mv ~/.config.pytm ~/.config.pytm.test
-	@mv ./.config.pytm ./.config.pytm.test
 	@python3 -m unittest
-	@mv ~/.config.pytm.test ~/.config.pytm
-	@mv ./.config.pytm.test ./.config.pytm
 
 .PHONY: describe
 describe:
--- a/README.md
+++ b/README.md
@@ -59,6 +59,15 @@ Element class attributes:
   onAWS                           default: False
 
 ```
+When excluding threats, you can use a config file to list the ones to be excluded:
+
+```
+[Default]
+exclude = THREAT1,THREAT2,THREAT3,...
+```
+
+The file can be at your home directory (~/.config.pytm), at the same directory of the script (./.config.pytm or ./<scriptname>.pytm). Over time we will add more configuration options in the file.
+
 
 ## Model
 
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -7,6 +7,7 @@ import os
 import random
 import sys
 import uuid
+import configparser
 from collections import defaultdict
 from collections.abc import Iterable
 from enum import Enum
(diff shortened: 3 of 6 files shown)


pytm/pytm.py
```diff
@@ -418,6 +419,23 @@ def _get_elements_and_boundaries(flows):
     return (list(elements), list(boundaries))
 
 
+def _load_config(tm):
```
nineinchnick: This should be a method of the TM class

pytm/pytm.py
```diff
@@ -418,6 +419,23 @@ def _get_elements_and_boundaries(flows):
...
+    home_config = os.path.join(os.path.expanduser("~"), ".config.pytm")
+    global_config = os.path.join(os.path.dirname(__file__), "config.pytm")
+    tm_config = sys.argv[0].replace(".py", ".pytm")
+    filenames = [global_config, home_config, tm_config, ".config.pytm"]
+
+    config = configparser.ConfigParser()
+    config.read(filenames)
+    if config.sections() is []:
+        print("\nno config file found\n")
```
nineinchnick: You should use the logger here

Makefile
```diff
@@ -45,7 +45,11 @@ report: $(models:.py=/report.html) seq dfd
 
 .PHONY: test
 test:
+	@mv ~/.config.pytm ~/.config.pytm.test
```
nineinchnick: Tests should not modify files in the home dir

pytm/pytm.py
```diff
@@ -554,13 +572,17 @@ def __init__(self, name, **kwargs):
...
+        _load_config(self)
+
     @classmethod
     def reset(cls):
         cls._flows = []
         cls._elements = []
         cls._threats = []
         cls._boundaries = []
         cls._data = []
+        cls._threatsExcluded = []
```
nineinchnick, line 569: It might be out of scope of this PR but we should stop using class level attributes. This one doesn't have _any_ reason to be such and could be converted to a regular attribute.

pytm/pytm.py
izar: I was sure you mentioned before to *not* add stuff like this to TM for some reason or another...

pytm/pytm.py
izar: debugging, will remove.

Makefile
izar: The problem is that if there is a config file that it will break the JSON test, so I am moving it away then moving it back after the testing. The net effect is zero.

nineinchnick: It means the code is not easily testable. The default file location should be saved somewhere in an attribute to be modified during tests to point to a file located in the test dir.

pytm/pytm.py
nineinchnick: I don't recall. If you'd make this a method you wouldn't have to modify any global vars (the TM class) and the first argument would become `self`. So it's _almost_ a method already.

pytm/pytm.py
izar: let's call for a cleanup version then. we stop adding new stuff and just clean up.

pytm/pytm.py
izar: done.

pytm/pytm.py
izar: removed.

Makefile
izar: created a _ignore_config attribute that defaults to False, and set it in the json tests. But I have no clue why the json tests fail now, the output is completely different from the input.json (I added the new attribute to input.json).

pytm/pytm.py
```diff
@@ -573,7 +557,7 @@ def __init__(self, name, **kwargs):
         random.seed(0)
 
         # load the config file
-        _load_config(self)
+        TM._load_config(self)
```
nineinchnick: Don't pass self if you call this as a method.

README.md
```diff
@@ -59,6 +59,15 @@ Element class attributes:
...
 
 ```
+When excluding threats, you can use a config file to list the ones to be excluded:
+
+```
+[Default]
+exclude = THREAT1,THREAT2,THREAT3,...
+```
+
+The file can be at your home directory (~/.config.pytm), at the same directory of the script (./.config.pytm or ./<scriptname>.pytm). Over time we will add more configuration options in the file.
```
nineinchnick, line 69: I don't like the idea of having configs in multiple locations that can affect the generated report contents. There's already a way to exclude particular threats by setting an attribute in the TM class. This allows to have "reproducible builds" and is very explicit and always gets committed together with the model. Having a config in my home dir means any other collaborator will generate a different report. At work we even commit all generated reports and have a pipeline job that checks if it can generate same reports to make sure they're up to date.

pytm/pytm.py
```diff
@@ -786,17 +797,16 @@ def process(self):
             with open(result.json, "w", encoding="utf8") as f:
                 json.dump(self, f, default=to_serializable)
 
-        if result.report is not None:
-            print(self.report(result.report))
-
-        if result.exclude is not None:
-            TM._threatsExcluded = result.exclude.split(",")
```
nineinchnick, line 793: As per my other comment, maybe this shouldn't be a "private" attribute and be properly documented to encourage usage.

README.md
izar: damn, completely forgot we had the attribute way to exclude threats. We need better documentation stat :)

izar:
Forgot that there is an attribute providing the same functionality.

PR #137, opened 2020-12-13
allow overriding Finding attrs

nineinchnick:
Allow overriding Finding attributes in particular data flows (or other assets). For example, this allows to set custom responses, CVSS scores (added as new attrs) or increase/lower the severity. See the new test for usage examples.

If there's an override defined for a particular threat, it'll always be added as a finding, even if the threat no longer applies. This is to include all mitigations in a report. To remove the finding, remove the override.

Overrides are defined per element because otherwise the original threat should be adjusted.

Overrides can be defined in an asset and then they'd apply to all dataflows. For more granularity, they should be defined only in particular dataflows. There's a check to avoid having two overrides for the same threat, since it would be hard to debug why an override doesn't get applied when it would be overwritten by another one.


izar:
How about adding an example on readme ?

nineinchnick:
> 
> 
> How about adding an example on readme ?

Great idea, done! I also added a bit more info in the usage section.

izar:
Good stuff!

PR #138, opened 2020-12-28
Updating the README file

chadeckles:
Updating the README with the following information:
 •	Including an Introduction subheading that briefly explains the “why” of this project. 
•	Inclusion of a Features section.
•	Breaking out a "Getting Started" and "Usage" section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,10 +5,16 @@
 
 ![pytm logo](docs/pytm-logo.svg)
 
+## Introduction
 
-Define your system in Python using the elements and properties described in the pytm framework.
-Based on your definition, pytm can generate, a Data Flow Diagram (DFD), a Sequence Diagram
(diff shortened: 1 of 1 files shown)


nineinchnick (approved):
This is great!

nineinchnick:
You did mention on Slack that the threats section is confusing. I agree and I think you could:
* remove the `Currently supported threats` section and replace it with a reference to https://github.com/izar/pytm/blob/master/docs/threats.md - note that in the report templates I used I was creating references to particular threats in a file like this (in my own repo)
* start the `Threats database` section with a sentence explaining that there's a default threat database provided but it may generate lots of false positives, depending on the domain of the system being modeled

But this seems like out of scope of this PR so we can do it in another one, so treat this only as a suggestion.

izar (approved):

PR #139, opened 2021-01-25
Fix typo in DS06 which made it always True

eskilandreen:
I took the liberty of adding a test too. Feel free to disregard it if you don't think it's necessary.

### DIFF
--- a/pytm/threatlib/threats.json
+++ b/pytm/threatlib/threats.json
@@ -1306,7 +1306,7 @@
     "Likelihood Of Attack": "High",
     "severity": "Very High",
     "prerequisites": "",
-    "condition": "target.hasDataLeaks",
+    "condition": "target.hasDataLeaks()",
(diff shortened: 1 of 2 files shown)


tests/test_threats.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class TestThreats(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        with open(TM.threatsFile.default, "r", encoding="utf8") as threat_file:
+            threats_json = json.load(threat_file)
+        cls.threats = {x["SID"]: Threat(**x) for x in threats_json}
+
+    def test_ds06(self):
```
nineinchnick: All threat tests are in another existing file. Can you add this test after https://github.com/izar/pytm/blob/master/tests/test_pytmfunc.py#L884 ?

eskilandreen: Thanks for pointing that out @nineinchnick ! I moved the test there and made it proper instead of relying on a mock. More code but better coverage.

nineinchnick: This looks great, thanks!

nineinchnick (approved):

PR #141, opened 2021-03-15
Add missing _safeset to Data class

mikejarrett:
When attempting to generate dfd, seq or report with where the datastore has the attribute `isEncryptedAtRest` and data has `isStored` attribute I receive:

```
ERROR: test_dfd (tests.test_pytmfunc.TestTM)
----------------------------------------------------------------------
  File "/tmp/pytm/tests/test_pytmfunc.py", line 115, in test_dfd
    self.assertTrue(tm.check())
  File "/tmp/pytm/pytm/pytm.py", line 697, in check
    _apply_defaults(TM._flows, TM._data)
  File "/tmp/pytm/pytm/pytm.py", line 378, in _apply_defaults
    d._safeset("isDestEncryptedAtRest", e.sink.isEncryptedAtRest)
AttributeError: 'Data' object has no attribute '_safeset'
```

We're able to reproduce with this short example:
```
web = Server("Server")
db = Datastore("Db", isEncryptedAtRest=True)
data = Data("Data", isStored=True)
dataflow = Dataflow("Server to DB', data=[data])
```

I was unsure which test to extend and I am happy to move to a different one or write a new one.

This appears to be a result of https://github.com/izar/pytm/pull/127

I ran the following before committing:
```
make test
make format
```

### DIFF
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -1179,6 +1179,12 @@ If only derivative data is stored (a hash) it can be set to False.""",
     def __str__(self):
         return "{0}({1})".format(type(self).__name__, self.name)
 
+    def _safeset(self, attr, value):
+        try:
+            setattr(self, attr, value)
+        except ValueError:
+            pass
+
 
 class Asset(Element):
     """An asset with outgoing or incoming dataflows"""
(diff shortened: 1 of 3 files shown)


nineinchnick (approved):
LGTM, thanks! We missed this because `Data` does not inherit from `Element`.

PR #143, opened 2021-03-19
moved to DFD3: lambdas are not special symbols, and datastore labels …

izar:
…appear outside of shape


izar:
dfd3

pytm/pytm.py
```diff
@@ -1354,19 +1351,34 @@ def __init__(self, name, **kwargs):
...
+        levels = kwargs.get("levels", None)
+        if levels and not levels & self.levels:
+            return ""
+
+        return self._dfd_template().format(
+            uniq_name=self._uniq_name(),
+            label=self._label(),
+            color=self._color(),
+            shape=self._shape(),
+            image=os.path.join(os.path.dirname(__file__), "images", "datastore.png"),
```
nineinchnick, line 1376: This image would look better if the top and bottom would not touch, that is if it would be a bit taller.

nineinchnick (approved):
Yay! Less HTML tables in labels!

pytm/pytm.py
izar: sample

izar: oh just noticed the flow names are broken now. Dammit graphviz.

izar: sample

colesmj: @izar Where is the code that renders the arrows here? Is the shaded vs unshaded arrow heads the implementation of the initiator mark?

nineinchnick: It's a combination of the `arrowtail` and `dir` options. `arrowtail` is always set to `onormal` (open) but is not drawn by default, when `dir` is `forward`. `dir` is `both` if a request and response are merged together into a single flow.

izar:
Talked with @colesmj and made sure he ok's the merge.

PR #145, opened 2021-03-22
Added a .sourceCode attribute to Elements, and a check for the age di…

izar:
…fference (in days) between the source code and the model.

~/Src/pytm source_and_check* ❯ ./tm.py --stale 200                   
Checking for code 200 days older than this model.
Paths should be relative to the directory where the TM script is being run.

~/Src/pytm source_and_check* ❯ ./tm.py --stale 20                    
Checking for code 20 days older than this model.
Paths should be relative to the directory where the TM script is being run.

./pytm/json.py is 175 days older than this model.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -36,11 +36,6 @@ jobs:
         # a pull request then we can checkout the head.
         fetch-depth: 2
 
-    # If this run was triggered by a pull request event, then checkout
-    # the head of the pull request instead of the merge commit.
-    # - run: git checkout HEAD^2
-      # if: ${{ github.event_name == 'pull_request' }}
-
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
       uses: github/codeql-action/init@v1
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -38,8 +38,8 @@ jobs:
 
     # If this run was triggered by a pull request event, then checkout
     # the head of the pull request instead of the merge commit.
-    - run: git checkout HEAD^2
-      if: ${{ github.event_name == 'pull_request' }}
+    # - run: git checkout HEAD^2
+      # if: ${{ github.event_name == 'pull_request' }}
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
--- a/README.md
+++ b/README.md
@@ -79,12 +79,12 @@ optional arguments:
   --levels LEVELS [LEVELS ...]
                         Select levels to be drawn in the threat model (int
                         separated by comma).
-  --stale STALE         checks if the delta between the TM script and the
+  --stale_days STALE_DAYS    checks if the delta between the TM script and the
                         code described by it is bigger than the specified
                         value in days
 ```
 
-The *stale* argument tries to determine how far apart in days the model script (which you are writing) is from the code that implements the system being modeled. Ideally, they should be pretty close in most cases of an actively developed system. You can run this sporadically to measure the pulse of your project and the 'freshness' of your threat model.
+The *stale_days* argument tries to determine how far apart in days the model script (which you are writing) is from the code that implements the system being modeled. Ideally, they should be pretty close in most cases of an actively developed system. You can run this periodically to measure the pulse of your project and the 'freshness' of your threat model.
 
 Currently available elements are: TM, Element, Server, ExternalEntity, Datastore, Actor, Process, SetOfProcesses, Dataflow, Boundary and Lambda.
 
--- a/README.md
+++ b/README.md
@@ -58,21 +58,33 @@ make
 All available arguments:
 
 ```text
-tm.py [-h] [--debug] [--json] [--dfd] [--report REPORT] [--exclude EXCLUDE] [--seq] [--list] [--describe DESCRIBE] [--sqldump DBNAME]
+usage: tm.py [-h] [--sqldump SQLDUMP] [--debug] [--dfd] [--report REPORT]
+             [--exclude EXCLUDE] [--seq] [--list] [--describe DESCRIBE]
+             [--json JSON] [--levels LEVELS [LEVELS ...]] [--stale STALE]
 
 optional arguments:
-  -h, --help           show this help message and exit
-  --debug              print debug messages
-  --dfd                output DFD (default)
-  --report REPORT      output report using the named template file (sample template file is under docs/template.md)
-  --exclude EXCLUDE    specify threat IDs to be ignored
-  --seq                output sequential diagram
-  --list               list all available threats
-  --describe DESCRIBE  describe the properties available for a given element
-  --sqldump DBNAME     dumps all threat model elements and findings into the named sqlite file (erased if exists)
-  --json               output a JSON file
-
+  -h, --help            show this help message and exit
+  --sqldump SQLDUMP     dumps all threat model elements and findings into
+                        the named sqlite file (erased if exists)
+  --debug               print debug messages
+  --dfd                 output DFD
+  --report REPORT       output report using the named template file (sample
+                        template file is under docs/template.md)
+  --exclude EXCLUDE     specify threat IDs to be ignored
+  --seq                 output sequential diagram
+  --list                list all available threats
+  --describe DESCRIBE   describe the properties available for a given
+                        element
+  --json JSON           output a JSON file
+  --levels LEVELS [LEVELS ...]
+                        Select levels to be drawn in the threat model (int
+                        separated by comma).
+  --stale STALE         checks if the delta between the TM script and the
+                        code described by it is bigger than the specified
+                        value in days
 ```
+
+The *stale* argument tries to determine how far apart in days the model script (which you are writing) is from the code that implements the system being modeled. Ideally, they should be pretty close in most cases of an actively developed system. You can run this sporadically to measure the pulse of your project and the 'freshness' of your threat model.
 
 Currently available elements are: TM, Element, Server, ExternalEntity, Datastore, Actor, Process, SetOfProcesses, Dataflow, Boundary and Lambda.
 
@@ -122,6 +134,7 @@ user.inBoundary = User_Web
 web = Server("Web Server")
 web.OS = "CloudOS"
 web.isHardened = True
+web.sourceCode = "server/web.cc"
 
 db = Datastore("SQL Database (*)")
 db.OS = "CentOS"
@@ -129,6 +142,7 @@ db.isHardened = False
 db.inBoundary = Web_DB
 db.isSql = True
 db.inScope = False
+db.sourceCode = "model/schema.sql"
 
 my_lambda = Lambda("cleanDBevery6hours")
 my_lambda.hasAccessControl = True
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -76,7 +76,7 @@ class varString(var):
 
 class varStrings(var):
     def __set__(self, instance, value):
-        if not isinstance(value, Iterable) and not isinstance(value, str):
+        if not isinstance(value, Iterable) or isinstance(value, str):
             value = [value]
         for i, e in enumerate(value):
             if not isinstance(e, str):
@@ -928,9 +928,9 @@ a brief description of the system being modeled."""
 
     def _stale(self, days):
         try:
-            argv0path = os.path.dirname(sys.argv[0])
+            base_path = os.path.dirname(sys.argv[0])
             tm_mtime = datetime.fromtimestamp(
-                os.stat(os.path.dirname(sys.argv[0]) + f"/{sys.argv[0]}").st_mtime
+                os.stat(base_path + f"/{sys.argv[0]}").st_mtime
             )
         except os.error as err:
             sys.stderr.write(f"{sys.argv[0]} - {err}\n")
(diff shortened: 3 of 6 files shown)


pytm/pytm.py
```diff
@@ -979,6 +1008,9 @@ class Element:
 a custom response, CVSS score or override other attributes.""",
     )
     levels = varInts({0}, doc="List of levels (0, 1, 2, ...) to be drawn in the model.")
+    sourceCode = varString(
+        "", required=False, doc="Location of the code that describes this element."
```
nineinchnick: What if there are multiple files? At worst, there can be dependencies resolved by some package managers.

pytm/pytm.py
```diff
@@ -909,6 +911,33 @@ def process(self):
...
+        )
+
+        for e in TM._elements:
+            if e.sourceCode:
+                src_mtime = arrow.get(
+                    os.stat(os.path.dirname(sys.argv[0]) + f"/{e.sourceCode}").st_mtime
+                )
+                age = abs(src_mtime - tm_mtime).days
+                if (age) >= days:
+                    print(
```
nineinchnick, line 960: Should it change the exit code or be actionable in some other ways than just displaying text?

nineinchnick (commented):
I'm not sure the relationship between a model and the sources of a system is so simple. This will be confusing if we imply it's required to identify source files to complete a model. And the requirement to update either the model or sources should be verified during code review of either one.

pytm/pytm.py
izar: it is not a must. just a nice to have, to further push the idea of having the tm stored as part of the code it describes. it is not a problem if it is a loose coupling, it doesn't aim for complete semantic equivalence.

pytm/pytm.py
izar: nope. it is purely informational.

pytm/pytm.py
```diff
@@ -909,6 +911,33 @@ def process(self):
...
+                        os.path.dirname(sys.argv[0])
+                        + f"/{e.sourceCode}"
+                        + f" is {age} days older than this model."
```
nineinchnick: What about the opposite case when the model is older than the source code? Is that an issue that the model references code that's unchanged for a long time? Ideally, the model should have the highest possible coverage and there are always parts of the codebase that "just work" and don't require any changes. On the other hand, I, as a developer, would like to know when I updated the source code but forgot to update the model.

pytm/pytm.py
nineinchnick: WDYT about making it a one-to-many relationship? That is store this as a `varStrings`? You'd have to add it, but you can use `varInts` as an example. It would still allow setting a single string value and would silently convert it to a list, so the usage would remain simple but flexible.

README.md
```diff
@@ -58,22 +58,34 @@ make
...
+  --exclude EXCLUDE     specify threat IDs to be ignored
+  --seq                 output sequential diagram
+  --list                list all available threats
+  --describe DESCRIBE   describe the properties available for a given
+                        element
+  --json JSON           output a JSON file
+  --levels LEVELS [LEVELS ...]
+                        Select levels to be drawn in the threat model (int
+                        separated by comma).
+  --stale STALE         checks if the delta between the TM script and the
```
nineinchnick:
```suggestion
  --stale STALE_DAYS         checks if the delta of modification time between the TM script and the
```

pytm/pytm.py
```diff
@@ -909,6 +911,33 @@ def process(self):
...
 
+        if result.stale is not None:
+            print(self._stale(result.stale))
+
+    def _stale(self, days):
+        print(f"Checking for code {days} days older than this model.")
+        print(
+            "Paths should be relative to the directory where the TM script is being run.\n"
+        )
+        tm_mtime = arrow.get(
```
nineinchnick: Sorry for nitpicking, but do we really need a new dependency for a non-core feature like this? :-)

tests/test_pytmfunc.py
```diff
@@ -1312,3 +1312,7 @@ def test_DR01(self):
         insert.isEncrypted = False
         threat = threats["DR01"]
         self.assertTrue(threat.apply(insert))
+
+
+# There is no test for TM._stale() since the dates would be a running target -
+# just assume it works.
```
nineinchnick: Remove the part about assuming it works, since there's no way to guarantee it's true. The test could create a source files (added to `.gitignore`) and set it's time to something far into the future using:

```python
import os

os.utime(path_to_file, (access_time, modification_time))
```

Then you can assert that the diff is always greater than a few days.

pytm/pytm.py
nineinchnick: PS This reminds me about https://threatspec.org/ since it would be also possible to do as annotations in source code.

pytm/pytm.py
nineinchnick: I guess you could do:
```python
% cat test.py 
from datetime import datetime
from pathlib import Path

a = datetime.fromtimestamp(Path("tm.py").stat().st_mtime)
b = datetime.fromtimestamp(Path("setup.py").stat().st_mtime)

print((a - b).days)

% python3 test.py
76
```

tests/test_pytmfunc.py
izar: oh and of course I forgot to catch the exception if the file doesn't exist. doh.

pytm/pytm.py
izar: but arrow is sooooo cool :D

pytm/pytm.py
izar: I think I used abs() in the difference. If I did I'll just change the message, you make a good point. If not I'll add it.

pytm/pytm.py
izar: varStrings is a great idea. I'll go that way.

pytm/pytm.py
nineinchnick: Yes, you did, but I think you should have not.

pytm/pytm.py
nineinchnick: Probably :-) When we didn't have any dependencies initially I saw that as a big advantage. I still believe that having as few dependencies as possible is best.

izar: true. going back to less deps.

tests/test_pytmfunc.py
izar: fixed the non-catching os errors. I'll fix the tests later....yey for technical debt.

pytm/pytm.py
```diff
@@ -73,6 +74,18 @@ def __set__(self, instance, value):
         super().__set__(instance, value)
 
 
+class varStrings(var):
+    def __set__(self, instance, value):
+        if not isinstance(value, Iterable):
```
nineinchnick: A string is iterable, you need to check explicitly for this by using `if not isinstance(value, Iterable) and not isinstance(value, str)`.
```
In [2]: from collections.abc import Iterable

In [3]: isinstance("", Iterable)
Out[3]: True
```

tests/test_pytmfunc.py
nineinchnick: You can add tests later, but please remove the comment.

pytm/pytm.py
```diff
@@ -979,6 +1036,9 @@ class Element:
 a custom response, CVSS score or override other attributes.""",
     )
     levels = varInts({0}, doc="List of levels (0, 1, 2, ...) to be drawn in the model.")
+    sourceCode = varStrings(
```
nineinchnick: The name of this attribute suggests it contains the actual source code, not file names. Please change to `sourceFiles`

pytm/pytm.py
```diff
@@ -909,6 +923,49 @@ def process(self):
...
+                    elif age <= -days:
+                        print(
+                            f"Model script {sys.argv[0]}"
+                            + " is "
+                            + str(-1 * age)
+                            + " days newer than source code file "
+                            + os.path.dirname(sys.argv[0])
+                            + f"/{src}"
+                        )
+            except os.error as err:
```
nineinchnick: `try/except` blocks should be as small as possible. I assume this is for calling `os.stat`. If yes, please move it there.

pytm/pytm.py
```diff
@@ -909,6 +923,49 @@ def process(self):
...
+    def _stale(self, days):
+        print(f"Checking for code {days} days older than this model.")
+
+        tm_mtime = datetime.fromtimestamp(
+            os.stat(os.path.dirname(sys.argv[0]) + f"/{sys.argv[0]}").st_mtime
+        )
+
+        for e in TM._elements:
+            try:
+                if e.sourceCode == "":
```
nineinchnick: This check should not be necessary if the default value for `e.sourceCode` would be an empty list.

pytm/pytm.py
```diff
@@ -909,6 +923,49 @@ def process(self):
         if result.list is True:
             [print("{} - {}".format(t.id, t.description)) for t in TM._threats]
 
+        if result.stale_days is not None:
+            print(self._stale(result.stale_days))
+
+    def _stale(self, days):
+        print(f"Checking for code {days} days older than this model.")
```
nineinchnick, line 930: This is not true as you have two checks, if the model is older and if the source file is older.

pytm/pytm.py
```diff
@@ -909,6 +923,49 @@ def process(self):
...
+                        os.stat(os.path.dirname(sys.argv[0]) + f"/{src}").st_mtime
+                    )
+                    age = (src_mtime - tm_mtime).days
+
+                    # source code is older than model by more than the speficied delta
+                    if (age) >= days:
+                        print(
+                            os.path.dirname(sys.argv[0])
+                            + f"/{src}"
+                            + f" is {age} days older than this model."
```
nineinchnick, line 952: This message doesn't seem to be correct, isn't `sys.argv[0]` name of the TM script, not the source file?

pytm/pytm.py
```diff
@@ -909,6 +923,49 @@ def process(self):
...
+        )
+
+        for e in TM._elements:
+            try:
+                if e.sourceCode == "":
+                    continue
+
+                for src in e.sourceCode:
+                    src_mtime = datetime.fromtimestamp(
+                        os.stat(os.path.dirname(sys.argv[0]) + f"/{src}").st_mtime
```
nineinchnick: I don't like the assumption that all source files must be in the same dir as the threat model script. Either document it, or detect and allow using absolute paths.

pytm/pytm.py
izar: thanks!

pytm/pytm.py
izar: I didn't think the size of the block mattered. TIL.

pytm/pytm.py
izar: thanks!

pytm/pytm.py
izar: I'm open to be convinced otherwise, but my intent has always been that the model lives together with the code it represents. Meanwhile I've documented it in the variable doc.

pytm/pytm.py
```diff
@@ -73,6 +74,18 @@ def __set__(self, instance, value):
         super().__set__(instance, value)
 
 
+class varStrings(var):
+    def __set__(self, instance, value):
+        if not isinstance(value, Iterable) and not isinstance(value, str):
```
nineinchnick: Sorry, this should be `if not isinstance(value, Iterable) or isinstance(value, str):`. I have not payed enough attention previously.

pytm/pytm.py
```diff
@@ -909,6 +923,49 @@ def process(self):
         if result.list is True:
             [print("{} - {}".format(t.id, t.description)) for t in TM._threats]
 
+        if result.stale_days is not None:
+            print(self._stale(result.stale_days))
+
+    def _stale(self, days):
+        try:
+            argv0path = os.path.dirname(sys.argv[0])
```
nineinchnick: nit: how about using `base_path` instead of `argv0path`?

pytm/pytm.py
```diff
@@ -909,6 +923,49 @@ def process(self):
...
+            tm_mtime = datetime.fromtimestamp(
+                os.stat(os.path.dirname(sys.argv[0]) + f"/{sys.argv[0]}").st_mtime
```
nineinchnick: You can use `argv0path` (or `base_path`) here also.

pytm/pytm.py
```diff
@@ -909,6 +923,49 @@ def process(self):
...
+        for e in TM._elements:
+
+            for src in e.sourceFiles:
+                try:
+                    src_mtime = datetime.fromtimestamp(
+                        os.stat(argv0path + f"/{src}").st_mtime
+                    )
+                except os.error as err:
+                    sys.stderr.write(f"{sys.argv[0]} - {err}\n")
+                    sys.stderr.flush()
```
nineinchnick, line 951: Should there be a `continue` here? `src_mtime` might be unset.

pytm/pytm.py
```diff
@@ -909,6 +923,49 @@ def process(self):
...
+
+                age = (src_mtime - tm_mtime).days
+
+                # source code is older than model by more than the speficied delta
+                if (age) >= days:
+                    print(f"{argv0path}/{src} is {age} days older than this model.")
```
nineinchnick: This message is still wrong. To match the condition. it should say that the model is `days` older than the source file. Looks like we really need tests for this :-)

pytm/pytm.py
```diff
@@ -909,6 +923,49 @@ def process(self):
...
+                elif age <= -days:
+                    print(
+                        f"Model script {sys.argv[0]}"
+                        + " is "
```
nineinchnick: How about making it `is only X days newer` (added `only`), to match the condition?

Also, is it wrong that the model is about the same age as the source file? Shouldn't it only matter that it has been updated _after_ the source file?

.github/workflows/codeql-analysis.yml
```diff
@@ -38,8 +38,8 @@ jobs:
 
     # If this run was triggered by a pull request event, then checkout
     # the head of the pull request instead of the merge commit.
-    - run: git checkout HEAD^2
-      if: ${{ github.event_name == 'pull_request' }}
+    # - run: git checkout HEAD^2
```
nineinchnick: Why did you disable this?

izar: Got an email from deepcode that it was not necessary anymore.

pytm/pytm.py
izar: i asked myself but doubted myself as well :)

pytm/pytm.py
izar: 👍🏻

pytm/pytm.py
izar: 👍🏻

pytm/pytm.py
izar: true. oversight.

pytm/pytm.py
izar: *sigh* true

pytm/pytm.py
izar: 👍🏻

.github/workflows/codeql-analysis.yml
nineinchnick: So remove it

nineinchnick (approved):

.github/workflows/codeql-analysis.yml
izar: done

PR #146, opened 2021-03-30
Introduce TMSequenceConfiguration class to add functionality to PlantUML sequence generation

mikejarrett:
Attempting to implement some of my ideas from https://github.com/izar/pytm/issues/142 . This should be fully backwards compatible and self contained.

I'm not entirely happy with passing config attributes to the new `sequence_line` methods.

Example PlantUML output generated from the `tests/sequence_config.plantuml` file:

sequence_config


pytm/pytm.py
```diff
@@ -900,9 +1002,6 @@ def process(self):
         if result.report is not None:
             print(self.report(result.report))
 
-        if result.exclude is not None:
```
nineinchnick: Is this related to this PR or was committed by accident?

pytm/pytm.py
```diff
@@ -601,6 +604,90 @@ def __str__(self):
...
+        unique_colors = [e for e in value if not (e in seen or seen_add(e))]
+        super().__set__(instance, unique_colors)
+
+
+class varTMSequenceConfiguration(var):
+    def __set__(self, instance, value):
+        self.data[instance] = value
+
+
+class TMSequenceConfiguration:
```
nineinchnick: I'd drop the `TM` prefix. Abbreviations like that don't make it clearer what's the class purpose and `SequenceConfiguration` is fine.

pytm/pytm.py
```diff
@@ -601,6 +604,90 @@ def __str__(self):
...
+            value = [value]
+
+        for i, e in enumerate(value):
+            errors = []
+            if not isinstance(e, str):
+                raise ValueError(
+                    f"expecting a list of str, item number {i} is a {type(e)}"
+                )
+            elif not COLOR_REGEX.match(e):
+                raise ValueError(f"item {i} was not valid hex color code, received {e}")
```
nineinchnick: Can the error message be a little more informative, that is be specific what's the expected format?

pytm/pytm.py
```diff
@@ -601,6 +604,90 @@ def __str__(self):
...
+
+        # Remove duplicates but preserve order. See https://stackoverflow.com/a/480227 for more
+        # information on why we assign `seen.add` method to a variable (tl;dr it is for speed)
```
nineinchnick: Does speed matter here? How many colors do you expect on average? If it does not, I'd simplify this. But I appreciate the comment :-)

pytm/pytm.py
```diff
@@ -601,6 +604,90 @@ def __str__(self):
...
+        super().__set__(instance, unique_colors)
+
+
+class varTMSequenceConfiguration(var):
+    def __set__(self, instance, value):
+        self.data[instance] = value
+
+
+class TMSequenceConfiguration:
+    """Helper class for PlantUML sequence diagrams"""
```
nineinchnick: Can this be more descriptive? How about 
```suggestion
    """Allows configuring how PlantUML sequence diagrams are drawn"""
```

pytm/pytm.py
```diff
@@ -818,40 +913,47 @@ def dfd(self, **kwargs):
         )
 
     def _seq_template(self):
-        return """@startuml
+        return """@startuml{hide_unlinked}
```
nineinchnick: nit: add a newline before `{hide_unlinked}`, don't worry about it being empty

pytm/pytm.py
```diff
@@ -818,40 +913,47 @@ def dfd(self, **kwargs):
...
+                participants += "\n".join(values)
+                participants += "\nend box"
+            else:
+                participants += "\n".join(values)
+
+        messages = ""
         for e in TM._flows:
-            message = "{0} -> {1}: {2}".format(
-                e.source._uniq_name(), e.sink._uniq_name(), e.display_name()
+            messages += e.sequence_line(
```
nineinchnick: How about just passing the whole config?

nineinchnick (approved):
Looks great!

pytm/pytm.py
```diff
@@ -1547,6 +1653,30 @@ def hasDataLeaks(self):
             for d in self.data
         )
 
+    def sequence_line(self, sequence_config, *args, **kwargs):
```
nineinchnick: Thanks for passing the whole config, but I think having a positional arg here when the base class doesn't might cause issues. Can you get it from `kwargs`?

nineinchnick:
@izar can you take a look at this?

izar:
so, while i recognize the utility of the sequence diagram improvements, I am still struggling with seeing this as 1st class functionality. Would it make sense to group sequence diagram functionality/improvements behind a SequenceDiag (or some such) object that we could change easily if we were to move away from plantuml at some point?

nineinchnick:
> Would it make sense to group sequence diagram functionality/improvements behind a SequenceDiag (or some such) object that we could change easily if we were to move away from plantuml at some point?

I like that idea since I remember it was very confusing to me what's the relationship between both diagram types when I started with pytm. But we can do this in a follow-up PR and still get this one merged.

izar:
I would much prefer we don't incur the technical debt if we can avoid it. What do you think, @mikejarrett ?

nineinchnick:
> I would much prefer we don't incur the technical debt if we can avoid it. What do you think, @mikejarrett ?

I don't see this as technical debt. This PR adds configuration to an existing feature of drawing sequence diagrams and as I understood you, we just agreed to refactor it into another class. If we don't want to do this right now and don't want to remove sequence diagrams completely, there's no reason to hold this PR hostage or reject it.

mikejarrett:
@izar and @nineinchnick , apologies for lack of communication it has been a busy couple of weeks. I'll see if I can come up with a different implementation in the coming days.

izar:
No worries @mikejarrett , I moved this to v1.1.4 in a month. I think it is a good one, we just need to encapsulate it.

izar:
Hi @mikejarrett do you have time to try and encapsulate this functionality before we merge? Thanks!

PR #148, opened 2021-03-30
Create full Data object from Dataflow.data string

izar:
The bug report came from a user - strange data objects appeared in the dictionary. This fixes it by creating sane objects instead.


pytm/pytm.py
```diff
@@ -164,7 +164,13 @@ def __set__(self, instance, value):
 class varData(var):
     def __set__(self, instance, value):
         if isinstance(value, str):
-            value = [Data(value)]
+            value = [
+                Data(
+                    name="data from dataflow",
+                    description=value,
+                    classification=Classification.UNKNOWN,
```
nineinchnick, line 194: Maybe we should just change the default in the `Data` class?

pytm/pytm.py
```diff
@@ -164,7 +164,13 @@ def __set__(self, instance, value):
 class varData(var):
     def __set__(self, instance, value):
         if isinstance(value, str):
-            value = [Data(value)]
+            value = [
+                Data(
+                    name="data from dataflow",
```
nineinchnick: Can you expand the PR description a bit to explain why it's better to have constant names here?

Is it confusing to be able to define `data` in two ways, as `Dataflow` properties and as separate objects with `carriedBy` filled in? Maybe those two ways should not be mixed? Or maybe defining data as either strings or objects should not be mixed.

The default name here is not accurate, since `Data` can be also defined in assets, not only data flows.

nineinchnick (commented):
I'm still not sure what problem this addresses. We can continue discussing this on Slack.

pytm/pytm.py
izar: from PUBLIC to UNKNOWN ? I'm fine with that, as it flags in the report that something is missing.

pytm/pytm.py
izar: the constant here goes back to the previous comment - upon seeing "data from dataflow" here, the user would go back and refine the script with more details. It is more of a placeholder than a constant name.

Forgot about the assets, thanks!

nineinchnick: But what's wrong with having the data string as the name then?

izar: clarified message to something that needs fixing by the model author

README.md
```diff
@@ -141,7 +141,7 @@ my_lambda_to_db.dstPort = 3306
 user_to_web = Dataflow(user, web, "User enters comments (*)")
 user_to_web.protocol = "HTTP"
 user_to_web.dstPort = 80
-user_to_web.data = 'Comments in HTML or Markdown'
+user_to_web.data = Data(description='Comments in HTML or Markdown', classification=Classification.PUBLIC)
```
nineinchnick: Name is a required param in the Data class.

pytm/pytm.py
```diff
@@ -164,13 +164,19 @@ def __set__(self, instance, value):
 class varData(var):
     def __set__(self, instance, value):
         if isinstance(value, str):
-            value = [Data(value)]
+            value = [
+                Data(
+                    name="FIXME: Data defined as string in Element, use Data objects instead",
```
nineinchnick: I still don't get why the string value is not the name of the Data object but the description.

PR #150, opened 2021-04-02
Expose Actor and Asset lists in report template

nozmore:
I was trying to use the report framework to mimc existing TM output where annotations for components (Assets in pytm) are separate from dataflows. I couldn't do this currently as Assets were only included in Elements so I replicated what was done for boundaries, flows and elements for assets. 

Everything works locally in my manual tests, also updated json test data to reflect inclusion of the asset list and confirm all tests now pass.

One difference here is that the other TM._lists are actually used in the code for logic purposes, boundaries, flows, elements are looped thru in various points to process the TM. Currently this not needed for Assets but seems reasonable there may be a use case in the future. 

I implemented the same for Actors.

Alternatively if we don't want to include additional TM._lists the report issue could be addressed by filtering elements in the TM.report method. 

`assets = [ obj for obj in TM._elements if not isinstance(obj, Asset) ]

         data = {
             "tm": self,
             "dataflows": TM._flows,
             "threats": TM._threats,
             "findings": self.findings,
             "elements": TM._elements,
             "assets": assets,
             "boundaries": TM._boundaries,
             "data": TM._data,
`


nozmore:
If we want to proceed with this approach I can do the same for Actors in this PR, otherwise I can revert and just filter elements in the report method.

pytm/pytm.py
```diff
@@ -869,6 +871,7 @@ def report(self, template_path):
             "threats": TM._threats,
             "findings": self.findings,
             "elements": TM._elements,
```
nineinchnick, line 875: I wonder should we replace `_elements` with `_assets`. It's easy to combine both lists and it'll be more explicit. The word _elements_ is ambiguous anyway. @nozmore @izar WDYT?

nozmore: I don't have a strong opinion either way as for the existence of _elements in the code. Obviously I agree with using assets since I raised the PR : ) I would probably go a step further and remove elements from json output and as mentioned add an Actor list so that objects are only referenced once. You could hold off on that for a major version to avoid breaking changes or do so now but add a bool flag on TM to remove elements and maintain existing functionality.

nozmore:
I updated the PR with a similar change for Actors so its ready to merge if desired.

nineinchnick (approved):

izar (approved):

PR #154, opened 2021-04-06
Add unique Finding ID instead of re-using Threat ID

nozmore:
Currently when parsing JSON the Elements contain a list of findings which references only the Threat Id. Since a Threat and given ID can be assigned via a Finding to different components it is not possible to reference a specific finding.

This adds an incremental id to finding starting a 1. which can be used to retrieve the referenced finding.

### DIFF
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -532,7 +532,8 @@ class Finding:
     severity = varString("", required=True, doc="Threat severity")
     mitigations = varString("", required=True, doc="Threat mitigations")
     example = varString("", required=True, doc="Threat example")
-    id = varString("", required=True, doc="Threat ID")
+    id = varInt("", required=True, doc="Finding ID")
+    threat_id = varString("", required=True, doc="Threat ID")
     references = varString("", required=True, doc="Threat references")
     response = varString(
         "",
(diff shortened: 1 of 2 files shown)


pytm/pytm.py
```diff
@@ -664,24 +665,27 @@ def _add_threats(self):
...
+                override_ids -= set(f.threat_id for f in e.source.overrides + e.sink.overrides)
             except AttributeError:
                 pass
 
             for t in TM._threats:
                 if not t.apply(e) and t.id not in override_ids:
                     continue
-                f = Finding(e, threat=t)
+
+                finding_count += 1
```
nineinchnick, line 687: Using a counter like this will make these ids change with every change to the model that would affect matches, like adding/removing assets or even reordering them. Another option would be to create a composite id from the thread id and asset "order" property (which is very badly named and should be just an id). Yet another option is to hash some attributes, like "name" and type.

Anyway, separating finding id and threat id is a great improvement, thanks!

nineinchnick (approved):

pytm/pytm.py
izar: +1

nozmore: I thought about a more complicated Id logic to handle finding matching between runs. I didn't go that route because there will be edge conditions where matching will fail or cause collisions. I thought for now IDs would just be unique to the scan, if we built logic to input findings then re-run to model against a new threat lib and reporting on closed findings then Finding matching logic and consistency makes sense.

nineinchnick: Sounds good for now. We can figure out how to use a more stable id later on.

izar (approved):

PR #155, opened 2021-04-08
TemplateEngine improvements, updated template.md

nozmore:
I updated the template to showcase #150 and #154. Also expanded template_engine to support the `call` operator with methods that return a list and support to call methods within a report utility class. Updated template.md with examples of each.

Updated tests and ensured tests passed. Also updated report test to write `output_current.md` At some point tests could use this but for now can be used to quickly commit the new `expected` value when the report changes. Added output_current.md` to `.gitignore`

I also added an example in the template.md to get the parents of each Boundary, to see this I updated tm.py with a hierarchy of Boundaries. If we think this is too much for the intro tm.py I can back those changes out.


nozmore:
oh also added entries in the Data table for carriedBy and processedBy.

docs/template.md
```diff
@@ -21,36 +21,119 @@ Name|From|To |Data|Protocol|Port
 {dataflows:repeat:|{{item.name}}|{{item.source.name}}|{{item.sink.name}}|{{item.data}}|{{item.protocol}}|{{item.dstPort}}|
 }
 
+<br>
```
nineinchnick: I don't think this is necessary. The headers should delimit sections sufficiently.

docs/template.md
```diff
@@ -21,36 +21,119 @@ Name|From|To |Data|Protocol|Port
...
+
+## Boundaries 
+
+{boundaries:repeat:
+Element|{{item.name}}
+|:----|:----|
+Description|{{item.description}}|
+InScope|{{item.inScope}}|
+Parent|{{item:utils:getParentName}}|
+Parents|{{item.parents:call:{{{{item.name}}}}, }}|
```
nineinchnick: Can't this be a regular table, instead of having a separate table per boundary?

The headers should be human-friendly, that is `InScope` should have space in it.

How people reading the report can tell the difference between `Parent` and `Parents`? Should we only keep `Parents`? Do we need this at all?

docs/template.md
```diff
@@ -21,36 +21,119 @@ Name|From|To |Data|Protocol|Port
...
+## Assets 
+
+{assets:repeat:
+
+<br>
+<hr>
+
+Element|{{item.name}}
+|:----|:----|
+Description|{{item.description}}|
```
nineinchnick, line 73: Same here, why there's a separate table for each asset? This is inconsistent with existing items, like `Data Dictionary`.

docs/template.md
```diff
@@ -21,36 +21,119 @@ Name|From|To |Data|Protocol|Port
...
+<br>
+<hr>
+
+Dataflow|{{item.name}}
+|:----|:----|
+Description|{{item.description}}|
+InScope|{{item.inScope}}|
+Finding Count|{{item:utils:getFindingCount}}|
+
+###### Threats
```
nineinchnick, line 116: This is duplicated.

docs/template.md
```diff
@@ -21,36 +21,119 @@ Name|From|To |Data|Protocol|Port
...
   &nbsp;
   &nbsp;
   &emsp;
 </details>
-}|
+}}
+
+}
+
+&nbsp;
```
nineinchnick: Why is this needed?

pytm/report_util.py
```diff
@@ -0,0 +1,42 @@
+
+class ReportUtils:
+    @staticmethod
+    def getParentName(element):
```
nineinchnick, line 4: Those should be defined as `@property` in the `Boundary` class. Then you wouldn't need this class at all.

tests/output.md
```diff
@@ -26,19 +26,217 @@ Name|From|To |Data|Protocol|Port
...
+<hr>
+
+Element|Web Server
+|:----|:----|
+Description||
+InScope|True|
+Type|Server|
+Finding Count|0|
+
+###### Threats
```
nineinchnick: It should not be a section header unless there would also be one with the element name.

nineinchnick (changes requested):
I'm not convinced we need such a complicated template as an example. I believe it would confuse new users. We should send a clear message that the templating engine in `pytm` is very simple by design, and if it's not enough, users should override it and use a different one.

docs/template.md
```diff
@@ -21,36 +21,119 @@ Name|From|To |Data|Protocol|Port
...
+</details>
+}}
+}
+
+<br>
+<br>
+
+## Data Flows 
+
+{dataflows:repeat(e):
```
nineinchnick: What does `(e)` in `repeat(e)` do?

docs/template.md
```diff
@@ -21,36 +21,119 @@ Name|From|To |Data|Protocol|Port
...
-## Potential Threats
-
-&nbsp;
+## Actors
 &nbsp;
 
-|{findings:repeat:
+Name|Description|isAdmin
+|:----:|:--------:|:----:|
+{actors:repeat:|{{item.name}}|{{item.description}}|{{item.isAdmin}}|
```
nineinchnick: You're not reporting findings for actors?

docs/template.md
```diff
@@ -21,36 +21,119 @@ Name|From|To |Data|Protocol|Port
...
+{dataflows:repeat(e):
+
+<br>
+<hr>
+
+Dataflow|{{item.name}}
+|:----|:----|
+Description|{{item.description}}|
+InScope|{{item.inScope}}|
+Finding Count|{{item:utils:getFindingCount}}|
```
nineinchnick, line 114: How did you choose which properties to display? Why not all of them?

docs/template.md
```diff
@@ -21,36 +21,119 @@ Name|From|To |Data|Protocol|Port
...
+<br>
+
+## Data Flows 
+
+{dataflows:repeat(e):
+
+<br>
+<hr>
+
+Dataflow|{{item.name}}
```
nineinchnick: This should not be a table, but a section header. Data flows not in scope and without any findings should not be displayed.

docs/template.md
nozmore: I felt the without the report was visually cluttered. I removed them and used &nbsp;, and made things consistent between each section.

docs/template.md
nozmore: There are no threats for Actors so I didn't include them. I didn't want to add a section for them and give the impression that we could have identified threats but didn't find any by including a 0 count or empty section. 

I can added it with the  `if` operator so that way we don't have to touch this template when/if we add threats to actors.

docs/template.md
nozmore: It could be but I felt with a meaningful description and a list of Parents, add adding overrides and levels would make the table too long for the user and ugly on one-line in source.

InScope, done.

Immediate Parent and All Parents?

docs/template.md
nozmore: That was an attempt and altering the variable name but should have removed it, done.

docs/template.md
nozmore: There are unique tables for each dataflow with a Section header for dataflows.

I don't agree, if someone when thru the trouble of modeling them having them visible with their annotations is relevant to understanding the design. I have added if condition to only display the Findings only when findings exist.

docs/template.md
nozmore: idk, I cleaned up some of the excess &nbsp; and <br>

pytm/report_util.py
nozmore: Personally I think we should keep methods and properties in pytm classes to only be things required for the core logic of pytm. Things like this that are only for reporting and not useful anywhere else should be separate. There is no need anywhere to have a `getParentName` method except using this "template engine", in code you'd just call element.inBoundary.name.

If we think there is another need for any of these functions by all means we should move them in.

tests/output.md
nozmore: Unfortunately this is the only way in MD to control font size so this what I did. I changed to using bold and it looks fine.

pytm/report_util.py
nineinchnick: We use similar helper functions in threat expressions to keep them short. We also have some used in diagrams (display_name, label). I really don't see a problem with adding some more.

But I still think we should not add such a class at all. The template in which you've used this is supposed to be only an example, which imho is too complicated right now.

docs/template.md
nineinchnick: If `All Parents` are sorted, why we need `Immediate Parent` if it's included always as the first or last position?

nineinchnick:
@nozmore @izar WDYT about keeping two template examples? One simple, one more advanced?

pytm/report_util.py
nozmore: Ok I viewed this file differently I see this as the default report the system is able to produce and IMO should eventually be as complete as possible. Thinking about it now, its in `docs` with a sample tm, ok.

To address both we could have a simple template in `docs`, I can use the original report and either 1) use my version in docs with a different name, 2) put it in the root, 3) keep it internal to pytm and add functionality to generate this with an arg for the template isn't provided.

docs/template.md
nozmore: I didn't test or investigate it enough to determine if its consistently sorted. I originally added just Parent and thought I'd save the reader the need to look at all elements to figure out the parent+grandparents and just added the list of all parents.

I have no problem with this but if you feel strongly I can remove Parent and just leave Parents.

nozmore:
lol I put that in my reply to an earlier comment before reading this. Yes, I think we are viewing this report differently I would prefer to have a fairly complete report and then in `docs` have a simple version to go along with the `simple.tm`

nozmore:
Why did my tests fail here? Is this executing against the test files in master or in the branch, locally everything passes and I didn't touch DFD logic.

docs/template.md
nozmore: I originally removed Parents and left Parent but I put them back as Immediate Parent and All Parents so that I could have an examples of the templating features.

nozmore:
Sorry I disappeared, life got in the way. I fixed some issues and cleaned up the reports, merged in master. I think this is ready to be merged, well at least reviewed : )

raphaelahrens:
Hi,

first I like the change and I wanted to add my thoughts.
Might it be a better idea to have extra repos for reports?
Since a fancy report does not have to be part of the software.
In addition one could also add build instruction and files in the repos which would be solely for the report.
For example I wanted to checkout if I could make a pdf report  and if I find the time I thought about pushing it to a separate repo, which could then be linked to.
Also the users could fork repot to adjust it to their needs (people love company logos and names).

This does not really affect this PR but maybe future PRs which want to change the report template.

izar:
I like the idea of more complex reports but I think they should go into a directory, not a separate repo, lest the two repos get out of sync version-wise.

nineinchnick:
I'm still not convinced it's the right thing to do to extend this basic template engine. It's already non-trivial to write advanced templates in it. I vote for replacing it with jinja2.

nozmore:
> I'm still not convinced it's the right thing to do to extend this basic template engine. It's already non-trivial to write advanced templates in it. I vote for replacing it with jinja2.

If it wasn't already done I'd agree. Its done, functional, adds minimal changes to the engine, and moves pytm forward. Replacing with another library can still be done later or use something to render the JSON output.

This is my only open source project, how to tie breakers work?

nineinchnick:
> > I'm still not convinced it's the right thing to do to extend this basic template engine. It's already non-trivial to write advanced templates in it. I vote for replacing it with jinja2.
> 
> If it wasn't already done I'd agree. Its done, functional, adds minimal changes to the engine, and moves pytm forward. Replacing with another library can still be done later or use something to render the JSON output.
> 
> This is my only open source project, how to tie breakers work?

I don't agree that changes proposed here are minimal.
@izar has they final word here. I can disagree and commit.

izar:
Sorry for the late reply, I am a bit overwhelmed these days.
I am reluctant to move into something like jinja2, even though I like it -
I fear we "get married" to it moving forward, but it is a bit of an
unfounded worry. I can't put my finger on it, but something bothers me.
Here's what I propose - as @nozmore says, it is done. How about we move
forward with this patch, and on the next "the template engine doesn't do X"
we reassess - if jinja2 or some other engine do it and we find out the move
is "easy" we go that way, otherwise we continue with this implementation.
Acceptable compromise?

Btw, I don't have the final word. I value you guys' opinion very much and I
will never overrule a good argument, especially considering how much I
always learn from each one of you. We are a democratic team.

pytm/report_util.py
nozmore: Previously added a basic and advanced template which addresses some of the comments here. 

Since we have talked about eventually replacing this template system with something else I think using this class is a better option for logic added only due to limitations in this templating system and will make it easier to remove if/when we move to something more advanced, rather than cluttering up the various classes with property methods that won't be needed.

nozmore:
Sounds good to me. I don't have plans to add any additional logic to the templating system. That said I do see additions to either the various internal classes or the report_util class when data needs to be formatted in a way for reporting. 

For example the PR about custom properties using a dict, I have a method drafted locally that I will push to report_util once the template changes are merged that will enable looping over the dict and prevent KeyError exceptions when the key does not exist in the dict.

pytm/report_util.py
nozmore: Marking as resolved. I added separated out the templates as requested and responded to the comments about using the helper class.

PR #156, opened 2021-04-12
Fixing the testing samples to conform with DFD3

izar:
Errr....i forgot to fix the tests when changing to DFD3. All the html on the saved compares is not needed anymore.

### DIFF
--- a/tests/dfd.dot
+++ b/tests/dfd.dot
@@ -66,7 +66,7 @@ digraph tm {
             datastore_SQLDatabase_0291419f72 [
                 shape = none;
                 fixedsize = shape;
-                image = "/Users/itarandach/Src/pytm/pytm/images/datastore.png";
+                image = "/home/runner/work/pytm/pytm/pytm/images/datastore.png";
                 imagescale = true;
                 color = black;
                 fontcolor = black;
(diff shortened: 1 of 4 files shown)


tests/dfd.dot
```diff
@@ -62,22 +59,19 @@ digraph tm {
...
-                    </table>
-                >;
+                label = "Web Server";
+                margin = 0.02;
             ]
 
             datastore_SQLDatabase_0291419f72 [
                 shape = none;
+                fixedsize = shape;
+                image = "/Users/itarandach/Src/pytm/pytm/images/datastore.png";
```
nineinchnick: This will make it fail for anyone else. Can we make it a relative path?

nineinchnick (approved):

tests/dfd.dot
izar: Just figuring that out now, the hard way...

izar: suggestions? the path appears in 3 files that we compare output to, and it is hardcoded there.

nineinchnick: It's being set to an absolute path on line 1406 in pytm.py. This is probably required, especially when pytm will be installed using pip. So we'd need to fix it in the test, before comparing files.

izar: i'll do it when the file is read.

izar: holly molly i think it works now.

PR #157, opened 2021-04-13
Fix encoding errors due to compatibility between #151 and #154

nozmore:
With #154 the `Finding.id` value is not an `int` and `Threat.id` is a `str`, the common encoding method needs to handle `id` differently between the 2 object types.

### DIFF
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -508,6 +508,12 @@ to a boolean True or False""",
         self.example = kwargs.get("example", "")
         self.references = kwargs.get("references", "")
 
+    def _safeset(self, attr, value):
+        try:
+            setattr(self, attr, value)
+        except ValueError:
+            pass
+
     def __repr__(self):
         return "<{0}.{1}({2}) at {3}>".format(
             self.__module__, type(self).__name__, self.id, hex(id(self))
@@ -597,6 +603,12 @@ Can be one of:
 
         for k, v in kwargs.items():
             setattr(self, k, v)
+
+    def _safeset(self, attr, value):
+        try:
+            setattr(self, attr, value)
+        except ValueError:
+            pass
 
     def __repr__(self):
         return "<{0}.{1}({2}) at {3}>".format(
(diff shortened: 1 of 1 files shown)


pytm/pytm.py
```diff
@@ -1701,9 +1713,16 @@ def encode_threat_data(obj):
...
+            attrs.append("threat_id")
+
         for a in attrs:
             v = getattr(e, a)
-            setattr(t, a, html.escape(v))
+
+            if (isinstance(v, int)):
+                t._safeset(a, v)
+            else:
+                t._safeset(a, html.escape(v))
```
nineinchnick, line 1728: `_safeset` ignored errors from `var*` attributes when trying to assign a value where it has already been done once. It's only used when resolving the model to figure out defaults for attributes where users did not explicitly assigned values.

Here, we have full control over how Finding objects are created. It should not be necessary. Also, using it here won't have any effect if the values are already assigned.

nineinchnick: Would it be sufficient to cast all attributes to strings? They're only used in reports anyway.

izar: Do you mean html.escape(str(v)) ? That solves the int problem, but the escaping doesn't seem to be 'taking'.

nineinchnick: Is there a test that fails on master? I'll take a look.

nineinchnick: Ah we should just skip encoding ids all together, if they're ints. Remove `id` from `attrs`.

nineinchnick: `Finding.id` should be a string.

nineinchnick: Like so: https://github.com/izar/pytm/pull/158

nozmore: _safeset was used because due to deepcopy the values are already set, this is resetting the values to encoded versions and throwing ValueError when using setattr.

I thought about using a str but decided against it since the 'id' is an int so I wrote code to respect that. Making it a str will impact json output and cause unnecessary work for the consumer to re-cast it as an int.

izar (approved):
Tested & works.

PR #162, opened 2021-05-06
Added datastore.png to setup.py

raphaelahrens:
datstore.png was not part of the `package_data` field so it was not
installed.


nineinchnick (approved):

izar:
Thanks!

PR #170, opened 2021-09-02
An empty threat model with ignoreUnused throws an error

raphaelahrens:
If a model has only elements which will not be displayed because
ignoreUnused is set to True the function _sort_elements() fails.

This is caused by a call to max(orders.values())` which then is
`max([])` which is not a computable value.

The error is caused by a module like this or an empty model.

```
from pytm import (
    TM,
    Actor,
)

tm = TM("my test tm")
tm.ignoreUnused = True

actor = Actor('actor')

if __name__ == "__main__":
    tm.process()
```

### DIFF
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -318,6 +318,8 @@ def _sort(flows, addOrder=False):
 
 
 def _sort_elem(elements):
+    if len(elements) == 0:
+        return elements
     orders = {}
     for e in elements:
         try:


nineinchnick:
I'd say this should still raise an error, just with a more actionable message.

raphaelahrens:
I am  not sure if the check is really sufficient, maybe it can be done in the call to `max()` directly.
For example `max([-1]+orders.values())

raphaelahrens:
> I'd say this should still raise an error, just with a more actionable message.

Should an empty model also raise such an error like the following?

```
tm = TM("my test tm")
tm.ignoreUnused = True

if __name__ == "__main__":
    tm.process()
```


Because this model is not raising any errors.

```
tm = TM("my test tm")

if __name__ == "__main__":
    tm.process()
```

izar (approved):

izar:
Thanks!

PR #173, opened 2021-09-15
HTML escaping missed the 'target' field when cleaning Findings

### LINKED ISSUE #172: {{item.target}} not rendering to report's threats section
I'm using the included example `tm.py` with `docs/template.md` and getting a blank `<p>  </p>` as every threat's "Targeted Element" where `{{item.target}}` is defined in the template. I assume the example threat model is supposed to match some of the conditions defined in `pytm/threatlib/threats.json` and show the matching element name (the specific process, dataflow, etc) within this field? Is this a bug or am I misunderstanding the functionality?

Running as documented: `python3 ./tm.py --report docs/template.md`

izar:
Fixes #172

### DIFF
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -659,7 +659,7 @@ Can be one of:
         )
 
     def __str__(self):
-        return f"{self.target}: {self.description}\n{self.details}\n{self.severity}"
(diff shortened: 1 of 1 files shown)


izar:
Wait, failing tests.

izar:
That works now.

PR #174, opened 2021-09-17
Fix: Excluded threat IDs are ignored when using --exclude argument

jnk22:
Hi,

the `--exclude` argument does not have any effect on the output/reports.
I could not find any related issue/pull request so far.

Example:
`python tm.py --exclude DR01,DS06 --report template.md > report.md`

(This should remove the threats *Unprotected Sensitive Data* and *Data Leak* from the list of threats at the sample TM)

The attribute **_threatsExcluded** are set when parsing arguments, but were not used after that.
This was fixed with a simple if-statement when adding threats to the output/report.

However, this also required the argument parsing of excluded threats to be moved further up. Otherwise, reports/output would have been processed before threats are even added to the list of excluded threats as it seems.

### DIFF
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -739,6 +739,7 @@ with same properties, except name and notes""",
         cls._threats = []
         cls._boundaries = []
         cls._data = []
+        cls._threatsExcluded = []
 
     def _init_threats(self):
         TM._threats = []
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -773,6 +773,9 @@ with same properties, except name and notes""",
                 if not t.apply(e) and t.id not in override_ids:
                     continue
 
+                if t.id in TM._threatsExcluded:
+                    continue
+
                 finding_count += 1
                 f = Finding(e, id=str(finding_count), threat=t)
                 findings.append(f)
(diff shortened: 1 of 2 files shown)


izar:
Thanks!!
Funny, I could swear it was working as advertised!
Any chance you might add tests for the functionality so we can detect future breaks?

jnk22:
Yes, I'll add some tests :)

izar:
Thanks!

jnk22:
I have added a test case for excluded threat IDs.

This verifies that already excluded threats (i.e. threats that have been properly parsed with `TM.process()`) will not be in the findings.

This does not verify that the argument parsing of `--exclude [ID1],[ID2]` works though.
I couldn't find any other test where you explicitly test argument parsing.
If this is also required, I'll happily add some test for that as well!

If there is anything else about my test that you want me to rework, feel free to request some more changes :)

izar:
Awesome! Will you be participating in Hacktoberfest ? If so, we can wait to merge until next week so this counts towards your contributions. Let me know.

jnk22:
I have signed up for Hacktoberfest now and linked my GitHub account, thanks for reminding me!

I don't know if there is anything else I need to set up first, but you can just merge this now if you want.
If it counts towards my contributions that's cool, if not, that is totally okay as well :)

izar (approved):

PR #179, opened 2021-09-30
Add enum for DatastoreType used in Datastore objects, removed isSQL, …

nozmore:
…updated sample tm, and updated existing tests


nozmore:
Let me know if anyone has thoughts on values for DatastoreType enum.

tests/test_private_func.py
```diff
@@ -214,7 +216,7 @@ def test_defaults(self):
             {"target": func, "condition": "not any(target.inputs)"},
             {
                 "target": server,
-                "condition": "any(f.sink.oneOf(Datastore) and f.sink.isSQL "
+                "condition": "any(f.sink.oneOf(Datastore) and f.sink.type == DatastoreType.SQL "
```
nineinchnick, line 219: What if we'd use subclasses instead of a dedicated attribute? It looks weird to do two different kind of type assertion here, `oneOf()` and `type == Enum`, instead we could stick to only `oneOf()`. It also makes it easier to extend `pytm` by defining subclasses, as adding values to an existing Enum would require monkey-patching.

nozmore:
I realized the readme and docs have to get updated as well. I haven't looked into how the doc html is generated yet but will before merge.

nineinchnick:
> I haven't looked into how the doc html is generated yet but will before merge.

Just run `make docs`

tests/test_private_func.py
nozmore: Unless we think there would be specific annotations for the subclasses I'd rather leave them as Datastore with annotations for what it is .. but I guess I could subclass DatastoreType then you could oneOf the type and accomplish the same thing.

PR #186, opened 2022-01-27
Adding uniqueId to get static references for findings

per-oestergaard:
In my use case, I want to synchronize the findings with an external risk system. To be able to do so, I added these features -
1. Each element can be provided with a uniqueId. The idea is that these are specified when writing the model and thus static.
2. Findings will get a uniqueId which is a combination of the uniqueId and the findings's ID. In this way, these are also static and can be used as the primary key to the external risk system.
3. To make communication easier, the Element's uniqueId can be added to the name and thus become visible on the diagrams. This enables smoother dialog when talking about the output as talking about ABC becomes very precise.

I apologize upfront for these things -
- I only changed the pytm.py file to now. Any changes to README.md etc, I'll await your first feedback.
- I spend considerably time trying to get the value of the TM varStrings in the Element class. I could not find a way, so I added a global variable hack.
- I did not add any tests. Any advice on what tests I should add would be welcome.

### DIFF
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -1915,8 +1915,8 @@ def encode_element_threat_data(obj):
                v = getattr(o, a)
                if (type(v) is not list or (type(v) is list and len(v) != 0)):
                   c._safeset(a, v)
-
-       encoded_elements.append(c)
+                 
+       encoded_elements.append(c)    
 
     return encoded_elements
 
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -1320,7 +1320,7 @@ a custom response, CVSS score or override other attributes.""",
     def __init__(self, name, **kwargs):
         for key, value in kwargs.items():
             setattr(self, key, value)
-        self.name = nameUniqueIdFormat.format(name, self.uniqueId)
+        self.name = self.nameUniqueIdFormat.format(name, self.uniqueId)
         self.controls = Controls()
         self.uuid = uuid.UUID(int=random.getrandbits(128))
         self._is_drawn = False
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -751,6 +751,7 @@ with same properties, except name and notes""",
         required=False,
         doc="A list of assumptions about the design/model.",
     )
+    uniqueFindingIdFormat = varString("{0}-{1}", doc="Default formatting of the uniqueId of findings. Argument 0 is the uniqueId of the element the finding is related to and argument 1 is the id of the finding. E.g., if you elements is called E1 and the finding DE01, the uniqueId of the finding becomes E1-DE01. If you prefer say DE01:E1, specify the format as {1}:{0}")
 
     def __init__(self, name, **kwargs):
         for key, value in kwargs.items():
@@ -809,7 +810,8 @@ with same properties, except name and notes""",
                     continue
 
                 finding_count += 1
-                f = Finding(e, id=str(finding_count), threat=t)
+                f = Finding(e, id=str(finding_count), threat=t,
+                            uniqueId=self.uniqueFindingIdFormat.format(e.uniqueId, t.id))
                 logger.debug(f"new finding: {f}")
                 findings.append(f)
                 elements[e].append(f)
@@ -829,6 +831,8 @@ a brief description of the system being modeled."""
         self._check_duplicates(TM._flows)
 
         _apply_defaults(TM._flows, TM._data)
+
+        self._check_duplicate_uniqueId(TM._elements)
 
         for e in TM._elements:
             top = Counter(f.threat_id for f in e.overrides).most_common(1)
@@ -895,6 +899,21 @@ a brief description of the system being modeled."""
                         right,
                     )
                 )
+
+    def _check_duplicate_uniqueId(self, elements):
+        # using Counter.itervalues()
+        # to check all unique list elements
+        ids = set()
+        for e in elements:
+            if e.uniqueId != "":
+                if e.uniqueId in ids:
+                    raise ValueError(
+                        "The uniqueId '{}' is used more than once. As the uniqueId must be unique, please provide unique values".format(
+                            e.uniqueId
+                        )
+                    )
+                else:
+                    ids.add(e.uniqueId)
 
     def _dfd_template(self):
         return """digraph tm {{
@@ -1295,6 +1314,7 @@ a custom response, CVSS score or override other attributes.""",
         doc="Location of the source code that describes this element relative to the directory of the model script.",
     )
     controls = varControls(None)
+    uniqueId = varString("", doc="A unique ID combined with thread ID that gives a unique, stable finding key which in turn can be used for synchronizing the findings with external systems.")
 
     def __init__(self, name, **kwargs):
         for key, value in kwargs.items():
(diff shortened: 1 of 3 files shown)


ghost:
### CodeSee Review Map:
[Review these changes using an interactive CodeSee Map](https://app.codesee.io/r/reviews?pr=186&src=https%3A%2F%2Fgithub.com%2Fizar%2Fpytm)

Review [in an interactive map](https://app.codesee.io/r/reviews?pr=186&src=https%3A%2F%2Fgithub.com%2Fizar%2Fpytm)

View [more CodeSee Maps](https://app.codesee.io/r/maps?src=https%3A%2F%2Fgithub.com%2Fizar%2Fpytm)

#### Legend
CodeSee Map Legend

pytm/pytm.py
```diff
@@ -1890,8 +1915,8 @@ def encode_element_threat_data(obj):
                v = getattr(o, a)
                if (type(v) is not list or (type(v) is list and len(v) != 0)):
                   c._safeset(a, v)
-                 
-       encoded_elements.append(c)    
+
+       encoded_elements.append(c)
```
per-oestergaard, line 1915:
```suggestion
                 
       encoded_elements.append(c)    
```

pytm/pytm.py
```diff
@@ -1267,7 +1292,6 @@ def _safeset(self, attr, value):
 
 class Element:
     """A generic element"""
```
per-oestergaard, line 1289:
```suggestion
    """A generic element"""

```

per-oestergaard (commented):
Avoiding changes to irrelevant parts of the file owing to auto-format and a line deletion

izar:
Hey, thanks for the addition!

A suggestion on tests would be that the uniqueId stays constant between runs. You could create an external file with a set of uniqueIds, run a TM and see that the newly generated ones match with the ones in the file.

About the global variable, we are trying to stay away from those as much as possible. What was it you weren't able to access on Elements? Let's see if we can fix that.

per-oestergaard:
> A suggestion on tests would be that the uniqueId stays constant between runs. You could create an external file with a set of uniqueIds, run a TM and see that the newly generated ones match with the ones in the file.

I added test test_uniqueid_two_runs

per-oestergaard:
> About the global variable, we are trying to stay away from those as much as possible. What was it you weren't able to access on Elements? Let's see if we can fix that.

I definitely also want to avoid global contexts ;) The problem seems to stem from the "weird" (to be at least) way variables in TM are handled. Settings and getters seems to be used and the getter does not work globally. 

If I break in __init__ of Class Element -
```python
    def __init__(self, name, **kwargs):
        for key, value in kwargs.items():
            setattr(self, key, value)
        self.name = nameUniqueIdFormat_shadow.format(name, self.uniqueId)
        self.controls = Controls()
        self.uuid = uuid.UUID(int=random.getrandbits(128))
        self._is_drawn = False
        TM._elements.append(self)
```
at the `self.name` line. My debugger evaluates nameUniqueIdFormat_shadow as expected to a string but TM.nameUniqueIdFormat what looks like class that will not reveal its value -
```
> /home/vscode/.local/lib/python3.9/site-packages/pytm/pytm.py(1337)__init__()
-> self.name = nameUniqueIdFormat_shadow.format(name, self.uniqueId)
(Pdb) p nameUniqueIdFormat_shadow
'{1}:{0}'
(Pdb) p TM.nameUniqueIdFormat
<pytm.pytm.varString object at 0x7fb60697cd30>
(Pdb) 
```
I cannot make it return the value
```
 p TM.uniqueFindingIdFormat.data
<WeakKeyDictionary at 0x7fb60697cca0>
(Pdb) p TM.uniqueFindingIdFormat.data('value')
*** TypeError: 'WeakKeyDictionary' object is not callable
(Pdb) p TM.uniqueFindingIdFormat.data['value']
*** TypeError: cannot create weak reference to 'str' object
```
If I instead break in say check of class TM, I get the same error for the TM (static) reference; however, it works using self -
```
> /home/vscode/.local/lib/python3.9/site-packages/pytm/pytm.py(839)check()
-> if self.description is None:
(Pdb) p TM.nameUniqueIdFormat
<pytm.pytm.varString object at 0x7f64acb45d90>
(Pdb) p self.nameUniqueIdFormat
'{1}:{0}'
```

Any ideas?

izar:
If there's any confusion or error in the way varStrings works, it is probably due to my own mistake. It seems to be working everywhere else, what makes it unique to your case? I mean, not to be all "it works on my machine", which I am not, I just want to try and understand better the uniqueness of this case. I will try to run it later today (or this week) and see if I can figure something out. Perhaps varStrings needs an explicit __get__ ?

per-oestergaard:
The "only" special I see is that I try to access the value from another class. I do not see that anywhere else in the code.

per-oestergaard:
If I subclass Element of TM:
```python
class Element(TM):
```
I can access the values. However that have side-effects (like including too much in json dumps), so tests fails.
I cannot create a new instance of TM `tm = TM()` as it has required values and I cannot provide them as I cannot get TM.name (a loop).
I was also trying
```python
TM.nameUniqueIdFormat.__get__(TM.nameUniqueIdFormat, TM.nameUniqueIdFormat.default)
```
But that seems only to get the default value.

Last, I tried adding
```python
self.value = value
```
to the class `var` `__set__`. That works when I run the project, but tests fails with 
```
======================================================================
ERROR: test_uniqueid_two_runs (tests.test_pytmfunc.TestTM)
----------------------------------------------------------------------
  File "/mnt/c/Users/redacted/pytm/tests/test_pytmfunc.py", line 299, in test_uniqueid_two_runs
    internet = Boundary("Internet",uniqueId="B1")
  File "/mnt/c/Users/redacted/pytm/pytm/pytm.py", line 1792, in __init__
    super().__init__(name, **kwargs)
  File "/mnt/c/Users/redacted/pytm/pytm/pytm.py", line 1325, in __init__
    self.name = TM.nameUniqueIdFormat.value.format(name, self.uniqueId)
AttributeError: 'varString' object has no attribute 'value'

----------------------------------------------------------------------
```
Any ideas?

izar:
Well that goes beyond my Python-fu. @nineinchnick some insight perhaps?

nineinchnick:
There's an `order` field, which name is unfortunate because that's actually an ID. It can be assigned automatically, as an incrementing number, or manually. 

Can you check if that would solve your issue?

per-oestergaard:
> There's an `order` field, which name is unfortunate because that's actually an ID. It can be assigned automatically, as an incrementing number, or manually.
> 
> Can you check if that would solve your issue?

Hi @nineinchnick, pleased to meet your.

Do understand you question, you suggest that I use the order instead of adding the unique id? Perhaps that could be a solution to the uniqueness issue. However, I would still like to add the number to the drawings etc. so I would end up with the python issue anyway :'(

nineinchnick:
> > There's an `order` field, which name is unfortunate because that's actually an ID. It can be assigned automatically, as an incrementing number, or manually.
> > Can you check if that would solve your issue?
> 
> Hi @nineinchnick, pleased to meet your.
> 
> Do understand you question, you suggest that I use the order instead of adding the unique id? Perhaps that could be a solution to the uniqueness issue. However, I would still like to add the number to the drawings etc. so I would end up with the python issue anyway :'(

The finding holds a reference to both the threat and the element it was matched against, so you can access the threat id and the element id (order). You should be able to extend pytm to change the way it renders things. I'd be happy to help you with that, ping me on our [Slack](https://pytm-dev.slack.com/).

per-oestergaard:
A way (around the Python issue, but perhaps also generic) could be to update `display_name`. I can see that `order` is added for `Dataflow`. To be backwards compatible, that would require a new boolean property say `showOrder` can could be set on where wanted. Not sure this can be set to TM, but I will look at it.

per-oestergaard:
I have PR'ed another method #190 and thus closing this

PR #200, opened 2022-07-25
Small changes to the dfd output so as to be able to change the background color and set spline to ortho 

jamesfunnel:
Small changes to the dfd output so as to be able to change the background color and set spline to ortho via command line arguments.


izar:
Hi, thanks for the PR!
I am wondering if it makes sense to have these options as first-class command line switches. They seem to be more on the "dot" side of things than the pytm, and would be changes made to the generated dot file, rather than something functionally attached to threat modeling functionality. What's your take on this?

jamesfunnel:
> 

The pytm.py generates the structure and look of the graphviz dot graph. Since there is no intermediate step between the python objects and the graphviz dot graph generation, the modification of the graph colors etc needs to be changed in the methods of the classes themselves (if we aren't to parse the generated dot file and make the changes therein or mannually search/replace afterwards) and are now hard coded... hence I modified it to be quicker to change the look of the graphs.

izar:
That's fair, but if we take that concept further, it would create N number of possible "just this option" to modify the output. Instead, how about we added a capability to include a file with global graphviz directives, automatically ? so you could have "--dfd --include-graphviz-options" and read a file graphviz.options from the same directory that would be included into the output, without "poluting" the pyttm command line. Makes sense?

jamesfunnel:
> That's fair, but if we take that concept further, it would create N number of possible "just this option" to modify the output. Instead, how about we added a capability to include a file with global graphviz directives, automatically ? so you could have "--dfd --include-graphviz-options" and read a file graphviz.options from the same directory that would be included into the output, without "poluting" the pyttm command line. Makes sense?

Yes, I agree, that would be a nice solution.. using an external json file or something to set the graphviz-options. However, taking that another step would be to decouple the Threat Model classes from Graphviz altogether and let a Graphviz class decide how to render each object along with its connections. Thus, the pytm.py file would be split into three files, one for the main function and command line options, one for the Threat Model classes and another for the rendering stuff (such as Graphviz)... then also the pytm.py file would be kept clean from Graphviz dfd-creation code and rendering plugins could be handled separately. ... Having said that though.. I do like minimalistic approaches like the current one :)

izar:
This decoupling is something @nineinchnick, @colesmj and I have discussed a number of times - separating the "presentation" layer from the objects themselves - but we didn't manage to get time to do. Is that something you'd be interested in owning? This would greatly facilitate adding other presentation frameworks, like C4, etc.

jamesfunnel:
Ah ok, I see, sadly my time is a bit limited as well.... I really enjoy this framework and I'm using it quite a bit, so maybe I'll do some more changes in the future... but will probably be more like a "need to have" basis though rather than refactoring and whatnot.

izar:
Understood. I'll keep this PR open to serve as a reminder for me for the necessity to decouple things around. Please make sure to submit anything else you come up with, that's how we evolve. Thank you so much for your time!

jamesfunnel:
Now I extracted the added dfd command line options into a separate library file so as to minimize messing around with the pytm.py too much in the future. The original command line args are duplicated to allow for old models to work without code changes (and without runtime patching pytm.py).... maybe will remove that from pytm.py in the future though.

izar:
sorry, i totally missed the last update! will review asap.

izar:
totally fell from my radar. I still feel like this upgrades dot/graphviz to a first class citizen by linking pytm to it via cli arguments. I think that if someone cares so much about the colors they can change the output dot file, and the ortho layout is already a flag on dot command line, so we would be basically duplicating functionality.

pytm tries to follow the Unix philosophy where each tool does something well. It delegates the details of diagramming to those tools that do it best, creating a starting point for the diagram rather than "the whole thing".

PR #213, opened 2023-03-15
[FEAT] support to serialize Findings as JSON

mhnrv:
Added a couple of functions to serialize the Findings class objects as JSON representations.

### DIFF
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -711,9 +711,25 @@ Can be one of:
         return "<{0}.{1}({2}) at {3}>".format(
             self.__module__, type(self).__name__, self.id, hex(id(self))
         )
-
+    
+    def __iter__(self):
+        yield from {
+            "description": self.description,
+            "details": self.details,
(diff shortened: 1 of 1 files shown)


izar:
hi and thanks for the PR! How is this different from the current json output capability (./tm.py [...] --json JSON_FILE) ?

mhnrv:
I just tweaked the script to only show the threats by calling 

```python
from pytm.pytm import TM

tm = TM()
tm.resolve() # instead of tm.process()
for finding in tm.findings:
  print(finding.to_json())
```

This way, one can view the findings and the corresponding target instead of the entire model. That is just my opinion 😅

izar:
I'm a bit confused perhaps about the use case you have in mind. 
How would this be different from jq '.findings' on the file generated by --json ?

izar:
Hi, @mohan-ravichandran , can this be closed ?

izar:
No answer from author.

PR #215, opened 2023-05-30
Update main.yml

izar:
Moving up Python versions.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,7 +19,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -19,7 +19,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.9, 3.10, 3.11, 3.12]
```
colesmj, line 22: 3.12 is still in developer preview. Are you sure you want to add support here?

izar:
ignore for a bit - i'm testing to see what's breaking. 3.12 won't be supported :)

PR #216, opened 2023-06-04
Create SECURITY.md

izar:


izar:
Must try again after changes to main.yml are merged.

SECURITY.md
```diff
@@ -0,0 +1,12 @@
...
+## Supported Versions
+
+| Version | Supported          |
+| ------- | ------------------ |
+| all     | :white_check_mark: |
+
+
+## Reporting a Vulnerability
+
+Please see CONTRIBUTING.md
```
colesmj, line 12: CONTRIBUTING does not say anything about vulnerabilities specifically (only bugs as defect types, or design/development issues). Rather than offloading to that page, SECURITY should make the vuln reporting process clear.
Do we have a GPG key for encrypting sensitive info in Issues (if someone wants to report a security vuln through Github issues, as directed by CONTRIBUTING)?

izar: no, but considering our use case, wouldn't that be a bit too much?

Issue #222, opened 2023-10-23
Cannot override findings, threats remain, DFD impacted, exception thrown for overrides len > 1


MichaelMcAleer:
Using the included [tm.py](https://github.com/izar/pytm/blob/master/tm.py) as an example here, not the code snippet from the README under '[creating a threat model](https://github.com/izar/pytm#creating-a-threat-model)'.

When I use `tm.py` to build an example report and DFD everything is fine. When I try to add a single override as demonstrated in the README section [overrides](https://github.com/izar/pytm#overrides) the threat is still listed under the dataflow for `user_to_web`, it is not overridden. 

The DFD generated from this threat model now has an `invalid` element because the `Finding` does not realise what element is referring to when it is initialised. Given the `Finding` is listed in the overrides for the dataflow it should know what element it is referring to.

![dfd](https://github.com/izar/pytm/assets/25844380/428dbf43-c4ed-4dd3-abfd-9d780324e715)

In addition to this, if an attempt is made to list more than one `Finding` in `overrides`, which expects a `list` of `Findings` in `varFindings()`, an exception is thrown:

```
❯ ./tm.py --report docs/advanced_template.md | pandoc -f markdown -t html > tm/report.html
Traceback (most recent call last):
  File "/Users/dev_user/PythonLibs/pytm/./tm.py", line 154, in <module>
    tm.process()
  File "/Users/dev_user/PythonLibs/pytm/pytm/pytm.py", line 1030, in process
    self.check()
  File "/Users/dev_user/PythonLibs/pytm/pytm/pytm.py", line 839, in check
    raise ValueError(
ValueError: Finding  have more than one override in Dataflow(User enters comments (*))
```

The example provided in the README uses threat SID `INP02` which isn't a threat identified for that dataflow, I have used `DE01` on its own to get the result above with invalid element in the DFD and remains as an active threat in the generated report. 

When overriding multiple threats I used threats identified in the report to ensure relevance and uniqueness. The `user_to_web` `overrides` is as follows in my altered `tm.py`:

```
user_to_web = Dataflow(user, web, "User enters comments (*)")
user_to_web.protocol = "HTTP"
user_to_web.dstPort = 80
user_to_web.data = comments_in_text
user_to_web.note = "This is a simple web app\nthat stores and retrieves user comments."
user_to_web.overrides = [
    Finding(
        id="DE01",
        CVSS="9.1",
        response="""**To Mitigate**: run a memory sanitizer to validate the binary""",
    ),
[...]

PR #223, opened 2023-10-24
Issue #222: fixed Findings creating spuriois Elements

izar:
Watching out for the "invalid" Element created when a Finding is added by hand.

### DIFF
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -242,14 +242,15 @@ class DataSet(set):
     def __str__(self):
         return ", ".join(sorted(set(d.name for d in self)))
 
+
 class varControls(var):
     def __set__(self, instance, value):
         if not isinstance(value, Controls):
             raise ValueError(
-                "expecting an Controls "
-                "value, got a {}".format(type(value))
+                "expecting an Controls " "value, got a {}".format(type(value))
             )
         super().__set__(instance, value)
+
 
 class Action(Enum):
     """Action taken when validating a threat model."""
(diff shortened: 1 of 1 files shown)


pytm/pytm.py
```diff
@@ -242,15 +242,16 @@ def __ne__(self, other):
...
         return ", ".join(sorted(set(d.name for d in self)))
 
+
 class varControls(var):
     def __set__(self, instance, value):
         if not isinstance(value, Controls):
             raise ValueError(
-                "expecting an Controls "
-                "value, got a {}".format(type(value))
+                "expecting an Controls " "value, got a {}".format(type(value))
```
nineinchnick, line 250:
```suggestion
                "expecting an Controls value, got a {}".format(type(value))
```

pytm/pytm.py
```diff
@@ -661,7 +667,7 @@ def __init__(
         if args:
             element = args[0]
         else:
-            element = kwargs.pop("element", Element("invalid"))
+            element = kwargs.pop("element", Element("created from Finding"))
```
nineinchnick, line 670: Maybe the `Element()` is used here as a default value only to avoid checking later if `element is None`. If that's the case, it's better to explicitly check for it in this function, than check for a reserved name elsewhere in the code.

pytm/pytm.py
```diff
@@ -442,18 +443,25 @@ def _apply_defaults(flows, data):
...
-                        d._safeset("isDestEncryptedAtRest", e.sink.controls.isEncryptedAtRest)
+                        d._safeset(
+                            "isDestEncryptedAtRest", e.sink.controls.isEncryptedAtRest
+                        )
                 if hasattr(e.source, "isEncryptedAtRest"):
                     for d in e.data:
                         d._safeset(
-                            "isSourceEncryptedAtRest", e.source.controls.isEncryptedAtRest
+                            "isSourceEncryptedAtRest",
+                            e.source.controls.isEncryptedAtRest,
```
colesmj, line 464: This change appears to go against the convention you have for other changes in this set - from a single line _safeset() to 2-line, but here you have 3-lines. Try to be consistent.

pytm/pytm.py
colesmj: In class Finding, element is set to required. If Finding can be connected to the tm rather than a specific Element, it seems this might address the problem. A dummy element is being created just to satisfy this constraint.

colesmj: Furthermore, Finding also has _2_ attributes on associating with Elements:

- element - required Element object
- target - optional - string?

Maybe we need to remove one?

colesmj (commented):
I think the change made - to change the name of the dummy finding on pytm.py:670 for easier filtering later is reasonable, but I think some more conversation is needed around Finding.element vs .target and whether .element needs to be required.

colesmj:
Trying for some PR cleanup. 
@izar or others is this still an issue that needs fixing?

izar:
Don't know, I'll test when I get home later today. 
Thanks for the cleanup!

izar:
invalid element resolved.

PR #225, opened 2023-10-28
Corrected the Overide example

raphaelahrens:
The example of the override feature has not been kept up to date and it include an error when setting the CVSS.


izar (approved):

izar:
Sorry it took so long - was away at ThreatModCon and OWASP DC.

PR #226, opened 2023-11-06
Added Error handling for User errors

raphaelahrens:
Some issues where raised where the users made an error when executing pytm on the shell.

It seems that the issue is that pytm is not returning a custom error but just a Python trace.

This commit add the UIError, which will be handled inside the process function.

UIError is a warpper arround another exception and adds a context description.

This commit only adds UIError for open calls in pytm, since these are obvious pitfalls for users.

### DIFF
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -34,6 +34,12 @@ from .template_engine import SuperFormatter
 """
 
 
+class UIError(Exception):
+    def __init__(self, e, context):
+        self.error = e
+        self.context = context
+
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


raphaelahrens:
I was also looking into ValueError which are a common source of errors.
But an ValueError can only be caught in the tm.py file.
So sadly it is not possible to replace these error with more user friendly message, without introducing boilerplate code

```
def tm_fun():
    web = Server("Web Server")
....
if __name__ == "__main__":
    tm.process(tm_fun)
```

Or to change the calling convention from 

```
python3 tm.py --report "$file"
```
to something like 
```
python3 -m pytm tm.py  --report "$file"
```
Since then you can catch the exceptions while importing `tm.py`


Either way I think it would help users to better distinguish between bugs and input errors.

izar (approved):

izar:
Thank you!

PR #228, opened 2023-12-11
README : updated the #creating-a-threat-model with an example of a Da…

FinestMaximus:
Updated README : updated the #creating-a-threat-model with an example of a Data Object


izar:
hey there, thanks a lot for the patch! 
can you check lines 167 and 168, it seems that you are using FALSE instead of False and that breaks it.

FinestMaximus:
my pleasure! and done for the update.

izar (approved):

PR #229, opened 2023-12-18
Colormap

izar:
Added --colormap as a companion to --dfd: paints elements green/yellow/red according to the severity of their findings.


colormap.dot
```diff
@@ -0,0 +1,134 @@
...
+            fontcolor = black;
+            style = dashed;
+            color = black;
+            label = <<i>Server/DB</i>>;
+        ]
+
+        datastore_SQLDatabase_f8af758679 [
+            shape = none;
+            fixedsize = shape;
+            image = "/Users/izar.tarandach/Src/pytm/pytm/images/datastore_gold.png";
```
colesmj: Do you want to hard-code this path into the source tree? Can they be relative paths instead?

pytm/pytm.py
```diff
@@ -1366,10 +1381,14 @@ def dfd(self, **kwargs):
         if levels and not levels & self.levels:
             return ""
 
+        color = self._color()
```
colesmj, line 1400: Is this setting the color of the entire dfd, or of a specific Element?

pytm/pytm.py
```diff
@@ -33,6 +33,17 @@
     By Chris Beaumont
 """
 
+def sev_to_color(sev):
```
colesmj, line 37: A general comment: How do you want to address setting the color for an Element when multiple Findings / Threats are associated with it where each has differing severities?

Also, when a Finding has a CVSS score, which translates to None, Low, Medium, High, Critical by the CVSS spec, can these be used instead?

pytm/pytm.py
```diff
@@ -33,6 +33,17 @@
...
+    # calculate the color depending on the severity
+    if sev == 5:
+        return "firebrick3"
```
colesmj: A comment on color choices - some people have a hard time with certain red/green color shades. Consider solid primary colors if available in dot.

colesmj: Also consider solid shading the shape rather than just the outline.

tests/output.json
```diff
@@ -1352,6 +1370,7 @@
             "protocol": "",
             "response": null,
             "responseTo": null,
+            "severity": 0,
```
colesmj, line 1373: Did you have any tests that used a non-zero severity? What about tests that tested the severity-setting logic on pytm.py:1498 (added)?

test.txt
raphaelahrens, line 1: Was this added by mistake?

foo.png
colesmj, line 1: This image seems to show firebrick shaded boundaries only - was this the intended effect? As a demo/test, it might be more effective to mix colors between elements and boundaries.

colormap.dot
izar: argh silly me

pytm/pytm.py
izar: element

pytm/pytm.py
izar: right now the final color of the element is the highest severity found among its findings, and right now i am ignoring the cvss score in favor of the severity hardcoded in the finding (from threats.json)

pytm/pytm.py
izar: the "full" (green, yellow, etc) colors look like someone went crazy in the 80's. Bright, impossible to read. So I went to a more subdued shade. 
Good call on the solid shading, let me see how that looks.

tests/output.json
izar: I did, but this specific one I think appears here because it is an element without a finding (so it defaults to 0)

test.txt
izar: most definitely

foo.png
izar: added by mistake

izar:
ready for re-review

tests/output.json
colesmj: Please check - all severity in this file are 0. Please be sure to add non-zero severity tests (you are using severity to determine color, correct?).

izar:
good catch.

colesmj:
The output.json still shows only 0 severity outcomes. I see your test code should generate High severity, did the output file not get updated?

izar:
The colormap testing is looking straight into the generated dot file, not on output.json - check dfd_colormap.dot, you'll see it has different colors compared with dfd.dot, those represent the severities.

colesmj (approved):

Issue #19, opened 2018-07-20
Logic is going to get messy fast, rules engine?


nozmore:
I've been thinking about this and relates to a few issues I've added recently.

I think the logic is going to get messy as we add more Threats, Mitigations, and add logic to alter severity while applying mitigations.

Does it make sense to continue creating a tightly coupled rules engine here vs using something existing? 

Idk what exists for Python. For Java I've worked with Drools that would be perfect for this. So much so I had the fleeting thought to port this to Java to use it.

izar:
Let's see if we can find something in python first. I am not familiar with Drools, what are its main characteristics that you find appropriate? It seems to be a business engine?
If at all moving from Python, let's go PROLOG.

colesmj:
I agree a Rules Engine will be necessary and important, if nothing else to manage the complexity as the threat list expands. We might consider a package like this: https://pypi.org/project/Intellect/

izar:
I agree, it will be necessary as we move forward and complexity increases.
I was looking at a Python-Prolog bridge, but this Intellect seems
interesting in the fact that it appears to take input via classes and
properties, just as we are structuring the input for pytm.

lfservin:
I'm looking into https://pypi.org/project/durable-rules/ and https://pypi.org/project/rule-engine/ . I just tried looking into Intellect, but I didn't really get the grip of it. Have you had any discussions on this topic since 2018? is there a preferred way to go? I'd try to re-create the same threats and come up with the same results. Does anyone have any other examples that could be run to try this approach?

izar:
We are currently testing a number of approaches - a "best" one has not been decided yet. So far we will progress with the simple boolean rules until we get to an approach that is worth moving to.

Thorsten-Sick:
I had good experience with plugin based models (Cuckoo sandbox signatures and most of the flexible components in PurpleDome https://github.com/avast/PurpleDome ).
Based in a Rule class most of the rules would be just a few lines with the option to put heaps of real code in there for identification of security flaws.

izar:
interesting, i had never seen PurpleDome, thanks!
Would you have an example of a Rule class translation of any of the rules we currently have?

Thorsten-Sick:
Purpledome was fun. But rule wise Cuckoo is more interesting. They use two system for rules. The newer one is "evented" because the logs they are processing are huge. Still they are quite simple to read. Examples are in https://github.com/cuckoosandbox/community/tree/master/modules/signatures/windows 

[...]

PR #231, opened 2023-12-21
Rule plugins simple2

Thorsten-Sick:
As discussed in issue #19 A PR for an additional plugin system to cover rule plugins. As an example it contains a semi complex sql injection plugin.


pytm/plugins/rules/specific_sql_injection.py
```diff
@@ -0,0 +1,61 @@
...
+
+    cwes = ["89", "1286"]
+    capecs = ["66"]
+    ttps = ["T1190"]
+
+    def __init__(self):
+        super().__init__()
+        self.plugin_path = __file__
+
+    def connected_elements(self, element):
```
colesmj, line 29: Could this be a function of an Element rather than in the rule definition? Seems it could be reused for other things, and taking a couple of extra parameters perhaps to be generically useful.

Thorsten-Sick: That would be better. I want to make changes by this feature in the core code as minimal as possible. Maybe create a separate PR for the Element upgrade ? Which parameters do you propose ?

Issue #234, opened 2024-03-09
Enhancement request: move pyDAL to local import


colesmj:
pyDAL is used for only 1 purpose within 1 function - pytm.py:1175 `SQLDump()`
For readability and to reduce pre-installation requirements for likely most users, consider making the import of pyDAL currently on pytm.py:24 be local to `SQLDump()`. This is allowed by Python and means if no one uses `SQLDump()` pyDAL is never loaded; otherwise it is a requirement and always loaded even if never used.

izar:
hm, but wouldn't that cause a fail if someone started slow and didn't use the sql functionality, but then did so? we would be banking on the idea that they know how to deal with an import error in Python.

raphaelahrens:
You can catch the failed import error and print guidance on how to proceed to stdout.

```
try:
    import pyDAL
except ImportError:
   print("please do ....")
```

izar:
... and I just learned a new one. This day already paid off, thanks!
In that case, PR away!

raphaelahrens:
I think this can be closed now.

PR #235, opened 2024-03-13
Added a first draft for #234

raphaelahrens:
In #234 @colesmj suggested to move the import of pydal into the sqlDumb function.
This commit does this and if the import fails raises an UIError with an explanation on how to proceed.
The text is just a first draft.

To move the import the function get_table was also moved inside the sqlDump function.

### DIFF
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -21,8 +21,6 @@ from textwrap import indent, wrap
 from weakref import WeakKeyDictionary
 from datetime import datetime
 
-from pydal import DAL, Field
-
 from .template_engine import SuperFormatter
 
 """ Helper functions """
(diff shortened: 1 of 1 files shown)


raphaelahrens:
Maybe you could add an explanation in the readme with instructions on how to proceed, which then can be linked in the error message.

izar:
Love the Freudian slip in the commit message :)
What happened to lines 1215-1223 ?

raphaelahrens:
> Love the Freudian slip in the commit message :) What happened to lines 1215-1223 ?

The function was moved from the the class into the method sqlDumP.
Since the get_table method is not using self and is only used inside of the sqlDump method. This was necessary because of the use of pydal.Field.

izar (approved):

PR #240, opened 2024-04-11
Revealjs & update scorecard action version

izar:
Added a template to create a https://reveal.js presentation.

**Needs tweaks**!

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -37,7 +37,7 @@ jobs:
           persist-credentials: false
 
       - name: "Run analysis"
-        uses: ossf/scorecard-action@99c53751e09b9529366343771cc321ec74e9bd3d # v2.0.6
+        uses: ossf/scorecard-action@v2.3.1
         with:
           results_file: results.sarif
           results_format: sarif
--- a/docs/reveal.md
+++ b/docs/reveal.md
@@ -1,0 +1,185 @@
+# {tm.name}
+
+---
+
+## System Description 
+
+{tm.description}
+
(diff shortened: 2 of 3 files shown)


colesmj:
Just a note - reveal.js is at https://revealjs.com/

.github/workflows/scorecard.yml
```diff
@@ -37,7 +37,7 @@ jobs:
           persist-credentials: false
 
       - name: "Run analysis"
-        uses: ossf/scorecard-action@99c53751e09b9529366343771cc321ec74e9bd3d # v2.0.6
+        uses: ossf/scorecard-action@v2.3.1
```
colesmj, line 40: Is this change related to the revealjs addition?

docs/reveal.md
colesmj, line 1: Do you have a sample of what this produces (maybe with the default threat model)?

izar: I'll add the movie to the readme - I don't want to pollute the tree with spurious files.

.github/workflows/scorecard.yml
izar: Let me remove this, it should be on an update of its own.

colesmj (approved):

colesmj:
@izar Did you plan to remove / revert the change to the ossf scorecard? If so I'll hold off on the merge so you can revert that change.

izar:
Nope, it's ok to go forward as is. Must have been some inconsistency between my local repo, most probably some mistake by me in git ops, but 2.3.1 is the right one.

Issue #237, opened 2024-03-29
Alternative format for threat lib for better readability and editing


raphaelahrens:
So I looked through the threats recently and again and again I find it difficult to read through the JSON document.

What bothers me the most is the combination of text in details, mitigation, and example with the metadata like id condtion severity and target.

It makes the text hard to read and the metadata difficult to find.

Now I had the idea to have a markdown file for each threat in the style of pandoc with the yaml_metadata_extension.

The result could look similar to this

```pandoc
---
id: INP01
condition: target.usesEnvironmentVariables is True and target.controls.sanitizesInput is False and target.controls.checksInputBounds is False
severity: High
target: (<class 'pytm.pytm.Lambda'>, <class 'pytm.pytm.Process'>)
---

# Buffer Overflow via Environment Variables

This attack pattern involves causing a buffer overflow through manipulation of environment variables. Once the attacker finds that they can modify an environment variable, they may try to overflow associated buffers. This attack leverages implicit trust often placed in environment variables.

## Example

Attack Example: Buffer Overflow in `$HOME` A buffer overflow in sccw allows local users to gain root access via the `$HOME` environmental variable. Attack Example: Buffer Overflow in TERM A buffer overflow in the rlogin program involves its consumption of the TERM environmental variable.

## Mitigations

Do not expose environment variable to the user.Do not use untrusted data in your environment variables. Use a language or compiler that performs automatic bounds checking. There are tools such as Sharefuzz [R.10.3] which is an environment variable fuzzer for Unix that support loading a shared library. You can use Sharefuzz to determine if you are exposing an environment variable vulnerable to buffer overflow.

## References

-  https://capec.mitre.org/data/definitions/10.html
-  CVE-1999-0906
-  CVE-1999-0046
-  http://cwe.mitre.org/data/definitions/120.html
-  http://cwe.mitre.org/data/definitions/119.html
-  http://cwe.mitre.org/data/definitions/680.html
```

Of course this file would not be parseable by pytm with the restriction that it should only depend on python stdlib.
But what could be done is that the threats are stored in markdown files from which the threatlib.json could be generated.

What do you think about this idea?

raphaelahrens:
Another advantage of this approach would be that code elements in the threats, could be marked as code and will not interfere with the formatting in the report.

An example of this is in SC04, which includes examples of jjencode

```js
[...]

PR #241, opened 2024-04-12
Added `prerequisites` and `likelihood` to Threat

raphaelahrens:
In threats.json the two properties ("prerequisites", "Likelihood Of Attack") are defined, but are not used in the rest of pytm.

This commit adds the two properties to the Threat class, so they can be used by other parts of pytm.

For me this was relevant, since I started to experiment with a different format for threats mentioned in #237 .
And after exporting threat.json to a markdown format and back into threat.json these two fields where missing.

### DIFF
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -605,8 +605,10 @@ class Threat:
 to a boolean True or False""",
     )
     details = varString("")
+    likelihood = varString("")
     severity = varString("")
     mitigations = varString("")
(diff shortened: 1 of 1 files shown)


izar (approved):

izar:
thanks!

raphaelahrens:
One question do you want to use the values somewhere? They could be used in the findings, but I still wonder why they where not in the Threat class and if there was a reason.

PR #242, opened 2024-04-18
Fixed #221 Got an error "AttributeError: 'str' ...

### LINKED ISSUE #221: Got an error "AttributeError: 'str' object has no attribute 'name'" when using the "--sqldump SQLDUMP".
I encountered the error:
```
dxu@remotedev-dxu-sea:~/open-source/pytm> ./tm.py --sqldump SQLDUMP
  File "./tm.py", line 127, in <module>
    tm.process()
  File "/ifs/home/dxu/open-source/pytm/pytm/pytm.py", line 1055, in process
    self.sqlDump(result.sqldump)
  File "/ifs/home/dxu/open-source/pytm/pytm/pytm.py", line 1149, in sqlDump
    for k, v in serialize(e).items():
  File "/ifs/home/dxu/open-source/pytm/pytm/pytm.py", line 1873, in serialize
    value = [v.id if isinstance(v, Finding) else v.name for v in value]
  File "/ifs/home/dxu/open-source/pytm/pytm/pytm.py", line 1873, in <listcomp>
    value = [v.id if isinstance(v, Finding) else v.name for v in value]
AttributeError: 'str' object has no attribute 'name'
```
Is it a mistake on my end, or could it be a bug?

raphaelahrens:
When pytm was run with the `--sqldump` flag with the example `tm.py` from the repository the execution failed with

```
AttributeError: 'str' object has no attribute 'name'"
```

This was caused by the `assumptions` attribute
https://github.com/izar/pytm/blob/6ca9f75ddaa5bda3503a6b8cbce5e6700e03e644/tm.py#L20-L22

When dumping the model into the database all attributes of the TM class are turned into strings, by first turning the obj into a dictionary, where specific attributes are removed and some are converted, and then each value in the dictionary are turned into strings.

This filtering and conversion is done by the `serilaize(obj, nested=False)` function. `sqlDump` transforms the values into strings.

The problem in #221 was that when `nested` is false the default behavior of `serialize()` is to assume that any list of values holds objects which have either a `.name` or are an instance of `Finding`. Since `assumptions` is a list of strings this fails.

The fix was to add `assumptions` to an already existing check for similar attributes.
Also the check was changed from `i == x or ...` to an `in` check.

But to be honest this code is very complex and holds many assumptions, which are not true for all classes and is constantly checking the type of the class.
Maybe it would be best to write specific serialize functions for some classes, and only have a genral serialize function which takes in an object and a blacklist of attributes.
The `to_serializable` singledispatch function already crates special functions for each class for the JSON conversion, maybe this can be extended.

### DIFF
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -1976,7 +1976,7 @@ def serialize(obj, nested=False):
                 value = value.name
             elif isinstance(obj, Threat) and i == "target":
                 value = [v.__name__ for v in value]
-            elif i == "levels" or i == "sourceFiles":
+            elif i in ("levels", "sourceFiles", "assumptions"):
                 value = list(value)
             elif (
                 not nested


izar (approved):
lgtm

izar:
Thanks, now I am wondering why this passed the tests so far.

raphaelahrens:
Good question, because the tests don't have an assumption in them. Should have checked the test.

raphaelahrens:
Another reason could be because there is no test for sql.

Issue #239, opened 2024-04-01
AC22 Credential Aging review


noloader:
Hi Everyone,

Threatlib has an item for AC22 Credential Aging:

```
{
    "SID": "AC22",
    "target": [
      "Dataflow"
    ],
    "description": "Credentials Aging",
    "details": "If no mechanism is in place for managing credentials (passwords and certificates) aging, users will have no incentive to update passwords or rotate certificates in a timely manner. Allowing password aging to occur unchecked or long certificate expiration dates can result in the possibility of diminished password integrity.",
    "Likelihood Of Attack": "Medium",
    "severity": "High",
    "prerequisites": "",
    "condition": "any(d.isCredentials for d in target.data) and target.sink.inScope and any(d.credentialsLife in (Lifetime.UNKNOWN, Lifetime.LONG, Lifetime.MANUAL, Lifetime.HARDCODED) for d in target.data)",
    "mitigations": "All passwords and other credentials should have a relatively short expiration date with a possibility to be revoked immediately under special circumstances.",
    "example": "",
    "references": "https://cwe.mitre.org/data/definitions/262.html, https://cwe.mitre.org/data/definitions/263.html, https://cwe.mitre.org/data/definitions/798.html"
  }
```

We have learned that continuity is a better security property than rotation. Unexpected changes, like gratuitously changing public keys changing, is bad for security because it breaks pinning controls. And requiring users to rotate their password results in users choosing weaker and weaker passwords over time just to comply with a policy based on reading tea leaves.

The details and mitigation detailed in threatlib are completely wrong nowadays. They run counter to what we have learned from real world incidents and security usability studies. Nowadays we want public keys and passwords written in stone, and only changed if there is suspicion or proof of breach or misuse.

I think AC22 should be either removed from the model or limited in scope.

izar:
Agreed. But being prepared for key rotation, on the other side, is an important requirement IMHO.
Would you like to submit a PR with better details and mitigation?

raphaelahrens:
What is the policy for changing an SID? Should this get a new SID or can this just be rewritten?

Further I would suggest to extract the `Lifetime.HARDCODED` condition and make this a new threat, since this is a design failure, which can not be fixed.

Here a  first draft to change AC22: 

# Credential disclosure

If credentials (passwords and certificates) have a long lifetime their disclosure can have severe consequences, if the credentials cannot quickly be revoked and/or rotated.

## Mitigations

[...]

PR #243, opened 2024-04-20
Replaced AC22 with AC23 and AC24

raphaelahrens:
As mentioned in #239 AC22 Credential Aging review the threat AC22 Credential Aging was not helpful.

This commit replaces AC22 with two new threats AC23 Credential Disclosure and AC24 Hardcoded Credentials.

AC23 checks if the lifetime of the credentials is LONG, MANAUL, or UNKNOWN.
Currently there is no way to resolve this threat by changing the model, besides setting the a different lifetime.

AC24 warns against the use of hardcoded credentials.

### DIFF
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -818,9 +818,9 @@ with same properties, except name and notes""",
             raise UIError(
                 e, f"while trying to open the the threat file ({self.threatsFile})."
             )
-
-        for i in threats_json:
-            TM._threats.append(Threat(**i))
+        active_threats = (threat for threat in threats_json if "DEPRECATED" not in threat)
+        for threat in active_threats:
+            TM._threats.append(Threat(**threat))
 
     def resolve(self):
         finding_count = 0
--- a/pytm/threatlib/threats.json
+++ b/pytm/threatlib/threats.json
@@ -1559,6 +1559,22 @@
     "references": "https://cwe.mitre.org/data/definitions/311.html, https://cwe.mitre.org/data/definitions/312.html, https://cwe.mitre.org/data/definitions/916.html, https://cwe.mitre.org/data/definitions/653.html"
   },
   {
+    "SID": "AC22",
+    "DEPRECATED": "AC22 was replaced by AC23 and AC24. Forcing short lived credentials by rotating the credentials regularly is no longer recommended. It better to go for long living strong credentials, which can be easily replaced, when a disclosure has happend.",
+    "target": [
+      "Dataflow"
+    ],
+    "description": "Credentials Aging",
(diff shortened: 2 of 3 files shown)


raphaelahrens:
@izar and @colesmj I would like your feedback on these two new threats. They are definitely not perfect, but I thought better to have a first draft.

Also I replaced AC22 since it is not clear how to proceed with old threats.

izar:
These look good to me, thanks! 
I wonder if just for the sake of completeness we shouldn't have an entry for AC22 saying it has been superseded by 23 and 24, without any rule, mostly for documentation-in-place sake. 
@colesmj ?

raphaelahrens:
@izar thanks. what about the fact "Currently there is no way to resolve this threat by changing the model, besides setting the a different lifetime.", meaning a shorter lifetime.

Is it ok to always get this finding, when using a large livetime or should there be a condition checking for a mitigation? Something like `has_high_entropy`?

izar:
I am not sure you can't mitigate it by changing the model. If it absolutely needs to have lifetime validity, it can be better protected, you can add "show current sessions" functionality, etc. 

Perhaps add  values LIFETIME_CANT_CHANGE and use that to point to additional controls, and create a finding for LIFETIME?

raphaelahrens:
>  I wonder if just for the sake of completeness we shouldn't have an entry for AC22 saying it has been superseded by 23 and 24, without any rule, mostly for documentation-in-place sake. @colesmj ?

Maybe we add a deprecated attribute to AC22 instead and when loading the JSON filter all threats with the deprecated attribute. The content of the attribute could also be the reason why it was detracted. The advantage would be that the old condition would still be intact.

Another idea would be to create a deprecated.json in add AC22 there.



> I am not sure you can't mitigate it by changing the model. If it absolutely needs to have lifetime validity, it can be better protected, you can add "show current sessions" functionality, etc.
> 
> Perhaps add values LIFETIME_CANT_CHANGE and use that to point to additional controls, and create a finding for LIFETIME?

Sorry I'm not sure what you mean. I see that you could add and use a custom value by adding

```
user_to_web.data = Data(
   "password", isCredentials=True, credentialsLife=random_value
)
```
Or do you mean that in pytm there should be a `Lifetime.CANT_CHANGE`?

izar:
> 
> Maybe we add a deprecated attribute to AC22 instead and when loading the JSON filter all threats with the deprecated attribute. 

Good idea!

> Or do you mean that in pytm there should be a `Lifetime.CANT_CHANGE`?

That.

raphaelahrens:
@izar Is there something missing, which I can do?

izar (approved):

izar:
Sorry, dropped this one  from my radar. Merged. Thanks!!

PR #245, opened 2024-05-28
Add Assumption to Elements

### LINKED ISSUE #244: Add a way to exclude threats for specific Elements
I have a situation where a lot of unrelated threats get detected. In this case the communication is happening via IPC, so I don't really need web relating threats, like PHP, XSS, XML etc.
However, this is only true for some parts of the system. There is also a web server and frontend involved, where some of these might apply. So I can't use the global `--exclude` flag. 

I could maybe modify the `threats.json` to check for a new custom property. 

But I think a more general solution might be, either, to modify `--exclude` to take an optional boundary name as a filter, so the exclude only applies within the boundary.

Or, in my opinion, a better solution would be to add an Element specific exclude. Similar to the overrides property. For example I (and probably others as well) have created custom classes, that inherit from the pytm classes and then inherit from each other. So this would also be a convenient way for excludes to be applied to multiple Elements. 

What do you think would fit better with pytm? Is the second solution fine, or does anyone have additional input?

If we decide on a specific implementation, I would be happy to create a pull request for the feature.

Dakes:
Fixes #244
This adds an Assumption class to Elements, as described in the Issue #244 by @raphaelahrens 

It allows someone to set a list of assumptions for elements, which also allow (optionally) to exclude a list of SIDs. 

All Findings, that match one of the SIDs, will not be included in the findings list. Instead there is a new list `excluded_findings` on the TM class. This allows the excluded threats to be shown in a separate section in the report. 

To demonstrate this, I also added this feature to the reference `tm.py` threat model and added a new section to the advanced template. 

I am not sure about the html files in the docs folder. I noticed there are make commands to update them, but they weren't updated in along while so I wasn't sure, whether to include them. But just let me know, if that was wrong and I will remove them from the Pull request.

### DIFF
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -218,8 +218,7 @@ class varData(var):
                 )
             ]
             sys.stderr.write(
-                "FIXME: a dataflow is using a string as the Data attribute. "
-                "This has been deprecated and Data objects should be created instead.\n"
+                "FIXME: a dataflow is using a string as the Data attribute. This has been deprecated and Data objects should be created instead.\n"
             )
 
         if not isinstance(value, Iterable):
@@ -227,7 +226,9 @@ class varData(var):
         for i, e in enumerate(value):
             if not isinstance(e, Data):
                 raise ValueError(
-                    f"expecting a list of pytm.Data, item number {i} is a {type(e)}"
+                    "expecting a list of pytm.Data, item number {} is a {}".format(
+                        i, type(e)
+                    )
                 )
         super().__set__(instance, DataSet(value))
 
@@ -852,7 +853,7 @@ with same properties, except name and notes""",
         findings = []
         excluded_findings = []
         elements = defaultdict(list)
-        for e in TM._elements:  # type: Element
+        for e in TM._elements:
             if not e.inScope:
                 continue
 
@@ -866,7 +867,7 @@ with same properties, except name and notes""",
             except AttributeError:
                 pass
 
-            for t in TM._threats:  # type: Threat
+            for t in TM._threats:
                 if not t.apply(e) and t.id not in override_ids:
                     continue
 
@@ -1374,7 +1375,7 @@ and only the user has), and inherence (something the user and only the user is).
 
 class Assumption:
     """
-    Assumption implemented by/on an Element.
+    Assumption used by an Element.
     Used to exclude threats on a per-element basis.
     """
     name = varString("", required=True)
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -271,6 +271,8 @@ class varControls(var):
 class varAssumptions(var):
     def __set__(self, instance, value):
         for i, e in enumerate(value):
+            if isinstance(e, str):
+                e = value[i] = Assumption(e)
             if not isinstance(e, Assumption):
                 raise ValueError(
                     f"expecting a list of Assumptions, item number {i} is a {type(e)}"
@@ -804,7 +806,7 @@ class TM:
         doc="""How to handle duplicate Dataflow
 with same properties, except name and notes""",
     )
-    assumptions = varStrings(
+    assumptions = varAssumptions(
         [],
         required=False,
         doc="A list of assumptions about the design/model.",
@@ -852,6 +854,8 @@ with same properties, except name and notes""",
         excluded_finding_count = 0
         findings = []
         excluded_findings = []
+        # We just need the assumptions with SIDs to exclude
+        global_assumptions = [a for a in self.assumptions if len(a.exclude) > 0]
         elements = defaultdict(list)
         for e in TM._elements:
             if not e.inScope:
@@ -875,7 +879,7 @@ with same properties, except name and notes""",
                     continue
 
                 _continue = False
-                for assumption in e.assumptions:  # type: Assumption
+                for assumption in e.assumptions + global_assumptions:  # type: Assumption
                     if t.id in assumption.exclude:
                         excluded_finding_count += 1
                         f = Finding(e, id=str(excluded_finding_count), threat=t, assumption=assumption)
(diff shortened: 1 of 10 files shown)


izar (approved):
Looks good to me!

izar:
This looks great, thanks! Can you please add tests to validate the new class ?

raphaelahrens:
@izar does this PR replace the excluded flag or do you see a value in having both?

izar:
@raphaelahrens really good question. I like this one because of the flexibility and cleanliess of the approach. Do you think we need both? Do you see a case to be made of one instead of the other?

pytm/pytm.py
```diff
@@ -840,19 +866,31 @@ def resolve(self):
...
+                for assumption in e.assumptions:  # type: Assumption
+                    if t.id in assumption.exclude:
+                        excluded_finding_count += 1
+                        f = Finding(e, id=str(excluded_finding_count), threat=t, assumption=assumption)
+                        excluded_findings.append(f)
+                        _continue = True
+                        break
+                if _continue:
+                    continue
+
```
raphaelahrens: This logic could be easier if the assumptions data structure would be a Dictionary<sid, [Assumption]> then you could just test `t.id in assumption_dict` and if so add the whole list of findings, which something like ` excluded_findings.extend((make_finding(a) for a in assumption_dict[i.id] ))
This would remove the inner loop and the `if _continue :`

pytm/pytm.py
```diff
@@ -824,9 +848,11 @@ def _add_threats(self):
 
     def resolve(self):
         finding_count = 0
+        excluded_finding_count = 0
         findings = []
+        excluded_findings = []
         elements = defaultdict(list)
-        for e in TM._elements:
+        for e in TM._elements:  # type: Element
```
raphaelahrens: the comment can be removed

pytm/pytm.py
```diff
@@ -218,17 +218,16 @@ def __set__(self, instance, value):
                 )
             ]
             sys.stderr.write(
-                "FIXME: a dataflow is using a string as the Data attribute. This has been deprecated and Data objects should be created instead.\n"
+                "FIXME: a dataflow is using a string as the Data attribute. "
```
raphaelahrens: Maybe revert this change, since the implicit string concat is confusing and this commit has nothing todo with this line.

pytm/pytm.py
```diff
@@ -1334,6 +1372,21 @@ def _safeset(self, attr, value):
             pass
 
 
+class Assumption:
+    """
+    Assumption implemented by/on an Element.
+    Used to exclude threats on a per-element basis.
+    """
+    name = varString("", required=True)
```
raphaelahrens, line 1385: What is the purpose of this name attribute? Isn't this the description?

raphaelahrens (commented):
Very nice.
I have some very minor comments and one about the name and description field in the Assumption class.

You added a name attribute and a description attribute, but in the `tm.py` example you only use the name attribute like a description. Maybe make the description mandatory and drop the name attribute.

raphaelahrens:
> @raphaelahrens really good question. I like this one because of the flexibility and cleanliess of the approach. Do you think we need both? Do you see a case to be made of one instead of the other?

Funny that you throw the ball back to me :laughing:. 
I think this solution makes the `--exclude` flag less useful, maybe useless.

`--exclude` is an all or nothing approach, and then the question is when do you need to exclude a whole threat.
The only thing for which it might be useful is when a threat is useless or at least the user considers it invalid.

Maybe in the case with AC22 it was useful, to exclude it while it was still in there.
But for pytm it was much better to create an issue and complain, than just to exclude it.

And as always there is backwards compatibility.

izar:
I think --exclude is still useful, as in cases where a mitigation is in
place that invalidates a class of threats.

pytm/pytm.py
raphaelahrens: Ok I looked into it and it is not really possible to define a `@property` for `self.assumption`.

The issue is that

1. the `varAssumptions` definition is overshadowed by the property
2. the `_safeset` method of `Element`is called multiple times and resets the values (really strange behavior)

So the current solution is fine.

pytm/pytm.py
Dakes: I used the name as a short description. (Similar to Dataflows)
The description could be optionally used, if someone wanted to elaborate on the Assumption. 
Because a longer description as the name could look weird, if you include it in a header, or preview in the report

pytm/pytm.py
Dakes: I agree, that the code isn't really nice. But if you say it's fine, I will leave it.
But I don't think making it a dict instead of list wold be good either way, because it would be more confusing for users to set. Also the SID would be duplicated as a key and inside the assumption.

Dakes:
tm.assumptions is currently a list of strings. Would it make sense to change it to be a list of `Assumption`? 
I think it would make sense, but would also break backwards compatibility. 
But it should be possible to make `varAssumptions`, to automatically create Assumptions from strings and return an Assumptions list. This would keep backwards compatibility. At least for the threat model code. Templates would still need adjusting.

raphaelahrens:
Mhh the logic of `tm.assumptions` would with the current logic not benefit from the change to a `list[Assumption]`.
Since there is no threat for the whole `TM`. For this to make sense it would need to integrate into the logic of `--exclude`
Or add extras logic for the `TM.assuptions`.

pytm/pytm.py
raphaelahrens: My idea was to define it in a list and only turning it into the dict when the model is processed/executed.But a proper label in python would be nice.

pytm/pytm.py
raphaelahrens: Still I would not call it name.
But lacking a better suggestion besides description/details and heading/descriptor I will shutup 😆

Dakes:
I now changed the Assumptions on the TM class to use the Assumption class in a way, that is 100% backwards compatible to existing threat models. It also will exclude threats globally, if it includes SIDs to exclude. (But still list them in the Excluded Section)

izar (approved):
Sorry for taking so long to get to this! Thank you for the PR!

PR #248, opened 2024-06-18
Fix cvss and response override

### LINKED ISSUE #222: Cannot override findings, threats remain, DFD impacted, exception thrown for overrides len > 1

Dakes:
This fixes the problem, that cvss and response by overrides aren't included in Findings. 

I think this also, at least partially, fixes #222

### DIFF
--- a/tests/test_pytmfunc.py
+++ b/tests/test_pytmfunc.py
@@ -294,7 +294,11 @@ class TestTM(unittest.TestCase):
         web = Server(
             "Web Server",
             overrides=[
-                Finding(threat_id="Server", response="mitigated by adding TLS"),
+                Finding(
+                    threat_id="Server",
+                    response="mitigated by adding TLS",
+                    cvss="1.234",
+                ),
             ],
         )
         db = Datastore(
@@ -304,6 +308,7 @@ class TestTM(unittest.TestCase):
                 Finding(
                     threat_id="Datastore",
                     response="accepted since inside the trust boundary",
+                    cvss="9.876",
                 ),
             ],
         )
@@ -328,8 +333,16 @@ class TestTM(unittest.TestCase):
             [f.response for f in web.findings], ["mitigated by adding TLS"]
         )
         self.assertEqual(
+            [f.cvss for f in web.findings],
+            ["1.234"],
+        )
+        self.assertEqual(
             [f.response for f in db.findings],
             ["accepted since inside the trust boundary"],
+        )
+        self.assertEqual(
+            [f.cvss for f in db.findings],
+            ["9.876"],
         )
 
     def test_json_dumps(self):
@@ -434,9 +447,6 @@ class TestTM(unittest.TestCase):
 
         self.assertTrue(tm.check())
         output = tm.report("docs/basic_template.md")
-
-        with open(os.path.join(output_path, "output_current.md"), "w") as x:
-            x.write(output)
 
         with open(os.path.join(output_path, "output_current.md"), "w") as x:
             x.write(output)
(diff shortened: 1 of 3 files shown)


izar (approved):
LGTM, thanks!

izar:
Can you add this case to the tests, before we merge? Thanks!

Dakes:
I now added a test for the private function encode_threat_data, because the alternative would be to test the generated reports. But I couldn't get the threats to show up in the report during the test. And it would also change the results if the threats.json would be updated. 

Also during writing the test I noticed another small bug, which I fixed in the last commit. If the inScope property of an Element was set to False, it would not reset the Findings, causing other tests to fail, besides the one, where the Finding was set. 
With these changes there is no code any more, which requires this fix, but I thought I'd add it anyway.

tests/test_pytmfunc.py
```diff
@@ -435,9 +448,6 @@ def test_report(self):
         self.assertTrue(tm.check())
         output = tm.report("docs/basic_template.md")
 
-        with open(os.path.join(output_path, "output_current.md"), "w") as x:
-            x.write(output)
-
```
Dakes, line 440: I removed this, because I assume, it is not intended to write the same file twice?

izar:
Thanks!

PR #249, opened 2024-06-20
version update

izar:


raphaelahrens:
Is it not 1.3.1?

izar:
damn. this is why i shouldn't do things in between things. thanks!

PR #252, opened 2024-07-26
Fixed json load for Controls and added test data to input.json

jmrenshaw:
Added controls to an example boundary in the input.json test file then resolved the resulting error where the class varControls doesn't recognise the dictionary provided by the input file.


izar:
superseded by pydantic rewrite

PR #259, opened 2024-12-11
Added flows module

raphaelahrens:
This new module introduces two new functions which help in the creation of dataflows.

The first function is `req_reply()`, which creates two dataflows. The first is usual dataflow and the second is a dataflow which is a response to the first.
I used this function a lot for modeling dataflow with databases.

The second functions `reply()` takes a dataflow as an argument and returns a new dataflow which is a response to the given dataflow. This function is quite useful for modeling proxies.

```python
d1 = Dataflow(client, web, 'Get page')
d2,d3 = req_reply(web, db, 'Query data')
d4 = reply(d1, 'return web page')
```

### DIFF
--- a/pytm/flows.py
+++ b/pytm/flows.py
@@ -1,0 +1,56 @@
+from pytm import Dataflow as DF
+from pytm import Element
+
+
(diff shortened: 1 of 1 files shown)


izar (approved):
Thanks!

PR #263, opened 2025-02-20
Multi threat libs

izar:
Adds capability to select which threat lib to use on the command line:

* nothing in the command line: uses the default library
* --threat-files foo.json : uses the threats in foo.json **only**
* --threat-files foo.json default : uses the threats in foo.json _and_ the default ones
*


tests/output.json
```diff
@@ -1405,5 +1405,5 @@
     "name": "my test tm",
     "onDuplicates": "Action.NO_ACTION",
     "threatsExcluded": [],
-    "threatsFile": "pytm/threatlib/threats.json"
+    "threatsFile": "{'pytm/threatlib/threats.json'}"
```
raphaelahrens, line 1408: Why is the output a string and not a list of strings?

pytm/pytm.py
```diff
@@ -838,16 +835,19 @@ def _init_threats(self):
...
-            raise UIError(
-                e, f"while trying to open the the threat file ({self.threatsFile})."
+
+        for tf in self.threatsFile:
+            try:
+                with open(tf, "r", encoding="utf8") as threat_file:
+                    threats_json = json.load(threat_file)
+            except (FileNotFoundError, PermissionError, IsADirectoryError) as e:
+                raise UIError(
+                    e, f"while trying to open the the threat file ({self.threatsFile})."
```
raphaelahrens: The error here should use `tf` to show which file was the one that was not accessible.

tests/output.json
izar: that's what varStrings with only one value put out...

pytm/pytm.py
izar: great catch, thanks!

izar: done

tests/output.json
raphaelahrens: Is this a bug?

izar: in the immortal words of just about everybody..."works for me"!

raphaelahrens: Ok I had a quick look and the issue is that the default `to_serializable` is used.
https://github.com/OWASP/pytm/blob/9dc0f1f04f8dc5ac0126b7950fb91fafc49efbd4/pytm/pytm.py#L2015-L2018

This is just the same as
```python
json.dumps(str(set(["./a.json"])))
```

The issue is created by 
https://github.com/OWASP/pytm/blob/9dc0f1f04f8dc5ac0126b7950fb91fafc49efbd4/pytm/pytm.py#L2023

Because of this check for `not nested` in `serialize()`.

https://github.com/OWASP/pytm/blob/9dc0f1f04f8dc5ac0126b7950fb91fafc49efbd4/pytm/pytm.py#L2064

https://github.com/OWASP/pytm/blob/9dc0f1f04f8dc5ac0126b7950fb91fafc49efbd4/pytm/pytm.py#L2035-L2070

This whole serialize function is a bit overloaded with special handling of member variables and might require a rewrite.
All the checks seem to be for specific classes which are all handle in the same function.
Also why is there the check for `nested`?
This is basically equivalent to checking if the class is `TM`.

A quick fix would be something like `if instance(obj, TM) and i == "threatsFile"` but oh boy that is not nice.

Should this be a new issue?

izar: sounds like something to be addressed. @nineinchnick , you there?

.gitignore
```diff
@@ -126,3 +126,4 @@ plantuml.jar
 tm/
 /sqldump
 /tests/.config.pytm
+.aider*
```
colesmj, line 129: @izar What is this for? What are `aider*` files?

README.md
```diff
@@ -60,10 +60,9 @@ make
 All available arguments:
 
 ```text
-usage: tm.py [-h] [--sqldump SQLDUMP] [--debug] [--dfd] [--report REPORT]
-             [--exclude EXCLUDE] [--seq] [--list] [--describe DESCRIBE]
-             [--list-elements] [--json JSON] [--levels LEVELS [LEVELS ...]]
-             [--stale_days STALE_DAYS]
+usage: tm.py [-h] [--sqldump SQLDUMP] [--debug] [--dfd] [--report REPORT] [--exclude EXCLUDE] [--seq] [--list] [--colormap]
```
colesmj, line 78: Are you adding `--colormap` in this PR also?

pytm/pytm.py
```diff
@@ -787,11 +788,7 @@ class TM:
     )
     name = varString("", required=True, doc="Model name")
     description = varString("", required=True, doc="Model description")
-    threatsFile = varString(
-        os.path.dirname(__file__) + "/threatlib/threats.json",
-        onSet=lambda i, v: i._init_threats(),
-        doc="JSON file with custom threats",
-    )
+    threatsFile = varStrings([_defaultThreatsFile])
```
colesmj, line 793: Minor nitpick - if this can ever be "one or more" threat files, consider threatsFiles (plural) for readability.

pytm/pytm.py
```diff
@@ -1129,6 +1129,21 @@ def _process(self):
         result = get_args()
         logging.basicConfig(level=logging.INFO, format="%(levelname)s: %(message)s")
 
+        # delaying loading of threats to accomodate multiple threat files in the
+        # command line
+        if result.threat_files:
```
colesmj, line 1136: Why do extra work on line 791 to load the default when you can just do it here? You will always reach this line at least once, correct?

tests/output.json
colesmj: There appears to be only 1 check for nested. If nested is true, the behavior seems to be potentially undefined (if the code reaches the `not nested` elif).

tests/test_pytmfunc.py
```diff
@@ -414,6 +414,23 @@ def test_json_loads(self):
             [f.name for f in tm._flows], ["Request", "Insert", "Select", "Response"]
         )
 
+    def test_threat_files(self):
+        dir_path = os.path.dirname(os.path.realpath(__file__))
+        foo_file = f"{dir_path}/1.json"
+        bar_file = f"{dir_path}/2.json"
+        threat_files = [os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
+    + "/pytm/threatlib/threats.json", foo_file, bar_file]
```
colesmj, line 399: Do you have a test for when the default is _not_ included in threatsfile?

tests/output.json
raphaelahrens: Yes. I started to rewrite the code see #268, but it is difficult to understand what the intention here was.
It seems that the default is `result[i.lstrip("_")] = value ` and the cases above only change the value of `value`.
So setting `nested = True` means that value will not be touched, except it is a `Element`, `Data`, or `Threat`. Or the name of the member is `in ("levels", "sourceFiles", "assumptions")`.

The last one seems to be a fix for a specific class.

.gitignore
izar: It's a co-pilot I have been testing. This is to make sure it doesn't get on git.

README.md
izar: it should be there already, it is an old one. It appears on the currently checked code on the master branch.

izar: oh this is the readme! let me fix that

pytm/pytm.py
izar: i don't remember the details but there is a timing issue there. Something around things being used before loaded.

tests/test_pytmfunc.py
izar: Hm. No. Gotta do one, you're right.

adaFPF:
Hi there!  I'd love to see this feature get accepted and ready for use downstream.  My understanding is that the following is what's holding it up:

1. #270 , which is in need of review, i.e. something I'm not in a position help with
2. Amending the test to handle the "no inclusion of default file" case

Anything else?

I can at least push an update to the test.  Would y'all prefer it directly pushed to this branch, or submitted as a separate PR to this branch?

izar:
@adaFPF sorry for the long silence - work overrode life. Pushing to this branch should be ok.

colesmj:
@izar I resolved the merge conflict on gitignore so this change can be merged.

colesmj:
@izar , @raphaelahrens is resolution for #268 + #270 a dependency for this (263)?

raphaelahrens:
> @izar , @raphaelahrens is resolution for #268 + #270 a dependency for this (263)?

There is the issue of the wrong serialisation.
https://github.com/OWASP/pytm/pull/263#discussion_r1988983800

But I would say #268  that this is not blocking this PR.

PR #269, opened 2025-03-23
Added Assumptions to the sql dump.

raphaelahrens:
The Assumptions table was not created if no assumptions where made, when running `python3 tm.py --sqldump`


raphaelahrens:
Sorry for the spam, but I misunderstod the code and this change is stupid.

izar (commented):
thanks!

Issue #278, opened 2025-10-07
Support ">" in Boundaries


fkromer:
If a [`Boundary`](https://github.com/OWASP/pytm/blob/master/pytm/pytm.py#L1942) contains `>` in it's name this results in a stdout/stderr output of  `Error: <stdin>: syntax error in line 27 near '\'` (line 27 is misleading).

PR #280, opened 2025-10-08
Fix for #278 missing html escaping for dot

raphaelahrens:
It was possible to inject html markup in the label of a dot node. This lead to the error observed in #278.

This fix is currently only for the label attribute. Other attribute might be affected as well.


izar:
I thought injecting HTML in the labels was a feature, not a bug, to give people more latitude in formatting their diagrams. Am I wrong?

raphaelahrens:
If it is a feature it needs to be documented. Since this was news to me.
But is this feature supported by all output formats in the same way?

The html in dot is just a reduced set of html and so has no security implications.
Maybe "inject" sounded a bit too much like a security problem :D, but it it most fitting to describe the bugin #278.

izar (approved):

PR #282, opened 2025-10-08
devbox: initial support

### LINKED ISSUE #281: Devbox as optional, convenient container alternative?
From a development convenience perspective isolating dependencies into a container is sub-optimal IMO. A more lightweight alternative for isolating development dependencies is [`devbox`](https://github.com/jetify-com/devbox) and results in better development convenience (e.g. the usual language server features when used in combination with an IDE of choice... I've tried with VSCode). What about adding `devbox` based dependency isolation as an alternative? @izar

fkromer:
closes #281

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -126,3 +126,6 @@ plantuml.jar
 tm/
 /sqldump
 /tests/.config.pytm
+
+# devbox
+devbox.lock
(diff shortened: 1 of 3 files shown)


izar (approved):

izar:
thanks, I learned a new thing today!

fkromer:
Same on my side :wink:

PR #288, opened 2025-10-09
editorconfig: init

### LINKED ISSUE #286: Provide editorconfig
If we'd provide a https://github.com/editorconfig (has support for a lot of IDEs) contributing would be eased.

fkromer:
Some first state as starting point.

closes #286

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,31 @@
+# EditorConfig helps maintain consistent coding styles for multiple developers working on the same project
+# More info: https://EditorConfig.org
+
+root = true
+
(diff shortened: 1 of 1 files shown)


fkromer:
Does not consider IDE specific configs like [jetbrains specific](https://www.jetbrains.com/help/pycharm/configuring-code-style.html#editorconfig) `ij_*` yet.

izar:
I need to say, I am a bit reticent to approve this one. I don't think we should be going out of our way to support every workflow and tool under the sun. Devbox made sense, but this one...not so much I think.

Issue #287, opened 2025-10-08
Modernize static code analysis


fkromer:
https://github.com/astral-sh/ruff is a generic replacement for a lot of [linters](https://docs.astral.sh/ruff/linter/). Provides a lot more than `pylint` and friends does. [Formats](https://docs.astral.sh/ruff/formatter/) as well (replaces `black`).

PR #290, opened 2025-10-11
add ruff

fkromer:


fkromer:
closes #287

fkromer:
The configuration defines defaults explicitly. Format rules are configured to work with google docstring format (has no [formal spec](https://realpython.com/documenting-python-code/#docstring-formats), is de-facto standard). Single quotes are migrated to double-quotes (cause de-facto standard.)

pyproject.toml
```diff
@@ -10,9 +10,58 @@ python = ">=3.9,<=3.11.13"
...
-black = "^25.9.0"
 pdoc3 = "^0.11.6"
+ruff = "^0.14.0"
 
 [build-system]
 requires = ["poetry>=0.12"]
 build-backend = "poetry.masonry.api"
+
+[tool.ruff]
+target-version = "py310"
```
fkromer, line 21: Support for 3.9 will run out end of October 2025. https://devguide.python.org/versions/

colesmj:
@fkromer Thank you for the requested changes, 

A couple of comments on your pull request:

1. This does not seem to be a clean PR for the intended purpose - the PR title is to add support for Ruff (a Rust-based linter for Python), yet there are multitude of other changes in your commits (mostly minor reformatting and other changes related to 3.9 end of support perhaps?). It would be ideal to have a clean PR for the addition of Ruff.

2. Ruff is a Rust-based linter for Python, a cutting edge module that is rapidly evolving - please make the case if you can why pylint, a mature python-based linter for python, is not sufficient or desired.

3. Thank you for the realpython link on docstrings. As a project we should consider which to standardize on; reStructuredText or NumPy/SciPy seem to be the more python-appropriate and formalized while the google docstring is simpler but as you highlighted there is not formal spec.

fkromer:
@colesmj Thanks for the review.

1. You are right, the minor reformatting is not specific to the PR. The purpose of the commit is to temporarily showcase in what code formatting would result in. I can revert the commits right before PR merge and provide as separate PR afterwards.
2. `ruff` is the de-factor standard in modern Python projects. It provides [hundreds of rules](https://docs.astral.sh/ruff/rules/) out of the box and replaces a variety of other tools (`black` -> formatting, `isort` -> import sorting, `mccabe` -> complexity analyzation, ...) cause `ruff` adopted their most reasonable and valuable rules. `fastapi` uses [`ruff`](https://github.com/fastapi/fastapi/blob/master/pyproject.toml#L211), `pydantic` uses [`ruff`](https://github.com/pydantic/pydantic/blob/main/pyproject.toml#L111), a lot of libraries in the ML domain use it. TBH I don't know of a single modern project which is using `pylint` these days anymore.
3. Right, it would make sense to agree on a docstring format still. Is not part of this PR. The current configuration has been made consistent with the google style docstrings which are in the current code base already. Changing those should be a rather simple AI driven "replace the google style docstrings with XXX style docstrings".

I'll comment in the PR for more fine-grained further explanation.

pytm/flows.py
```diff
@@ -3,7 +3,7 @@
 
 
 def req_reply(src: Element, dest: Element, req_name: str, reply_name=None) -> (DF, DF):
-    '''
+    """
```
fkromer, line 6: Kind of de-facto standard. I don't know any project which is using ```. Default formatter config value.

pyproject.toml
```diff
@@ -10,9 +10,58 @@ python = ">=3.9,<=3.11.13"
...
+extend-unsafe-fixes = []
+external = []
+fixable = ["ALL"]
+future-annotations = false
+ignore = []
+ignore-init-module-imports = true
+logger-objects = []
+per-file-ignores = {}
+preview = false
+select = ["E4", "E7", "E9", "F"]
```
fkromer, line 63: `ruff` rule set can be extended here. E.g. to add `pylint` rules refer to [here](https://docs.astral.sh/ruff/rules/#pylint-pl).

pytm/flows.py
```diff
@@ -22,17 +22,17 @@ def req_reply(src: Element, dest: Element, req_name: str, reply_name=None) -> (D
     Returns:
         a tuple of two dataflows, where the first is the request and the second is the reply.
 
-    '''
+    """
     if not reply_name:
-        reply_name = f'Reply to {req_name}'
+        reply_name = f"Reply to {req_name}"
```
fkromer, line 27: " is kind of-defacto standard (config default value).

pytm/json.py
```diff
@@ -4,17 +4,11 @@
 from .pytm import (
     TM,
     Boundary,
-    Element,
```
fkromer, line 7: Dead code not used, to showcase code cleanup capabilities (`--fix`).

pytm/template_engine.py
```diff
@@ -33,11 +32,11 @@ def format_field(self, value, spec):
             return result
 
         elif spec.startswith("call:"):
-           # Example usage, format, exampple format
```
fkromer, line 36: wrong comment indentation fixed.

tests/test_pytmfunc.py
```diff
@@ -4,7 +4,6 @@
 import re
 import unittest
 import tempfile
-from contextlib import redirect_stdout
```
fkromer, line 7: Dead code fixed (`--fix`).

fkromer:
Relates to #294

fkromer:
@colesmj

> Thank you for the realpython link on docstrings. As a project we should consider which to standardize on; reStructuredText or NumPy/SciPy seem to be the more python-appropriate and formalized while the google docstring is simpler but as you highlighted there is not formal spec.

I‘ve referenced some further examples w.r.t. docstring formats in #294 . Personally I’m totally fine with every alternative.

izar:
Superseded by #325 - if the docstrings need to change lets make a separate PR for that.

PR #299, opened 2025-10-29
add ruff gh workflow

### LINKED ISSUE #292: add ruff to gh workflows
relates to #287

fkromer:
- Closes #292 
- Extends #290 
- Action output will look similar [to this action output](https://github.com/fkromer/pytm/actions/runs/18922580892/job/54022234669) ("Check production code format" as well as "Check production code" will fail instead of pass like [here](https://github.com/OWASP/pytm/actions/runs/18922745884/job/54022800661?pr=299))

### DIFF
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -1,0 +1,48 @@
+name: ruff
+
+on:
+  push:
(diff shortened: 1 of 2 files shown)


izar:
Ruff is in already.

Issue #295, opened 2025-10-13
Use validation library for model input validation


fkromer:
The need for [validation functions](https://github.com/OWASP/pytm/blob/master/pytm/pytm.py#L89-L233) could be avoided if we'd define model classes as models using generic library support. These days the most common alternative for this purpose is [pydantic](https://github.com/pydantic/pydantic). `pydantic` enables:

- definition of domain [models](https://docs.pydantic.dev/latest/concepts/models/) 
- [type based model schema validation](https://docs.pydantic.dev/latest/why/)
- validation via built-int [field constraints](https://docs.pydantic.dev/latest/concepts/fields/#field-constraints)
- advanced validation via [Validators](https://docs.pydantic.dev/latest/concepts/validators/) and [Validation Decorators](https://docs.pydantic.dev/latest/concepts/validation_decorator/)

This would result in a rather bigger refactoring. Pls let me know what you think about this proposal @izar . I'd like to avoid the preparation of PRs which would not be accepted.

izar:
Moving to pydantic is in the roadmap, yes.

fkromer:
Support for [serialization](https://docs.pydantic.dev/latest/concepts/serialization/) could potentially be helpful (to resolve #268 ?) as well.

raphaelahrens:
The blocker for #268 is #273 since the sql dump code does not really make any sense to me and I don't know how this should be resolved.
The JSON serialization works.

Regarding the move to pydantic, I think it is a good decision.
But this would also require some form of packaging of pytm for different OSes.
Until now pytm had no dependencies besides python (except for the sqldump feature).
When using pydantic this would of course change.

colesmj:
> Regarding the move to pydantic, I think it is a good decision. But this would also require some form of packaging of pytm for different OSes. Until now pytm had no dependencies besides python (except for the sqldump feature). When using pydantic this would of course change.

Agreed pydantic would be the preferred choice - var-based enum-ing is very cumbersome now - although would introduce some dependency challenges as you highlight. But the benefit of readability and flexibility (especially if you want to do some fancy object customizations) may be significant to outweigh the challenge. Unlike the sql dependency checking to optionally enable the feature, this change for pydantic would be all or nothing (and a breaking change?).
Pydantic would not be the only Rust-based dependency in pytm (depending on outcomes of other PRs).

fkromer:
@raphaelahrens

[...]

PR #301, opened 2025-11-13
BREAKING CHANGE - removing sqldump - #295

izar:
As discussed, we are removing this functionality since I am the only one that needed it and now i dont and the code was junk.


NoodlesNZ:
@izar unless I'm missing something here, this is just documentation changes? No changes here to the code to remove this functionality

izar:
No review - documentation only

PR #313, opened 2026-01-16
Fix ERROR-level basedpyright type checking issues

shivachethanreddy:
This PR fixes ERROR-level issues reported by basedpyright / pyright.

Changes include:
- Fixing missing super().__init__() calls required by static analysis
- Correcting generic container typing
- Narrowing boolean handling so string → bool conversions are correctly understood
- Fixing container usage that caused type errors
- Updating pdoc_overrides to use an explicitly typed result (bool | str)

All fixes were verified locally using basedpyright.
Warnings were intentionally left out of scope.
No runtime behavior or API changes were made.

### DIFF
--- a/pytm/__init__.py
+++ b/pytm/__init__.py
@@ -24,7 +24,7 @@ __all__ = [
 ]
 
 import sys
-
+from typing import Dict, Union
 from .json import load, loads
 from .pytm import (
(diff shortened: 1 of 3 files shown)


izar:
Applied on #322, thanks!

Issue #12, opened 2018-07-19
Initial TM creation is a pain and requires a lot of typing


nozmore:
... well it was anyway.

https://github.com/izar/pytm/commit/9bccd8f45d5ce440b1f42f40ef8d62097cd45823

I added a python script to take a CSV with pairs of elements. I then create generic Element definitions for each unique name and create dataflow for each pair.

After editing the file to replace Element with Actor, Server, Process, etc I can generate a basic TM DFD then start to annotate each element and add boundaries as needed.

Before I do any more with this take a look and lets discuss. Initially I wanted the csv to be as lightweight as possible but we could have it contain various more data like variableName, displayName, element type, or various annotations. 

I've committed the geneate.py file, a sample csv, the generate sample.py and sample.png and then a modified (Element->Actor,Process, etc) py and png so you can see what its doing.

izar:
That's an interesting idea, but I have a bit of an allergy to CSV files due to them leading to Excel tomfoolery and, well - having CVS files that include stuff that eventually ends up in Python code. 

I've been battling excessive typing by using snippets on my editor.

nozmore:
You can say that with a straight face with the 'eval' calls in this code ; )

Could be another format, maybe JSON. As currently implemented (only variable names) it could have simple validation, likely just alphanumeric.

nozmore:
One of my submitted PRs has a modification which goes with this. I changed the name on Elements to the actual name instead of the hash of the name on the DFD and the color is red. I also made out of Scope elements a box instead of a circle.

With these changes looking at the DFD it is clear which items are still Elements.

izar:
Hey at least there's a caveat in the README about the eval, and threats.py is supposed to be more-or-less stable and write-protected...

Elements was supposed to be a parent class - not something to be represented. The Process is supposed to be the smallest granular "element" in use.

nozmore:
My thought was to use the Element as just a temporary state to when generating the base TM , the next step is to change all Elements to their appropriate type.

nozmore:
and just kidding about the eval, you are correct it is a _threat_ that needs to be addressed. : )

nineinchnick:
Since we already have json input, is this issue still valid? Can we close it?

cptfixit:
I'm pretty sure that, after two years of no response, this issue can be closed. 

I hope the JSON input is also documented. ;)

fkromer:
[...]

PR #314, opened 2026-01-28
Add JSON template and examples to simplify threat model creation

aryangupta060407-web:
### Summary
This PR adds a JSON template and an examples folder to simplify the initial creation of threat models in pytm.

### Motivation
Issue #12 mentions that creating a threat model requires a lot of manual typing.  
Providing a ready-to-use JSON template helps users quickly bootstrap their threat models with minimal effort.

### Changes
- Added `examples/template_tm.json` as a starter template for threat models.
- Updated README with instructions for using the JSON template.

### Related Issue
Addresses #12

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,12 +128,24 @@ Element class attributes:
   onAWS                           default: False
 
 ```
-## Quick Threat Model Creation Using JSON Template
+
+
+## Using the JSON Template
 
 To simplify initial threat model creation, pytm provides a JSON template.
 
-Users can copy `examples/template_tm.json` and modify it to define their threat model.
-This reduces the amount of manual typing required when creating a new threat model.
+1. Copy the file `examples/template_tm.json`.
+2. Modify the elements and flows according to your system.
+3. Load the JSON file in pytm using the built-in JSON import feature.
+
+Example:
+
+python
+from pytm import TM
+
+tm = TM("Example Threat Model")
+tm.load("template_tm.json")
+
 
 
 The *colormap* argument, used together with *dfd*, outputs a color-coded DFD where the elements are painted red, yellow or green depending on their risk level (as identified by running the rules).
--- a/README.md
+++ b/README.md
@@ -128,6 +128,13 @@ Element class attributes:
   onAWS                           default: False
 
 ```
+## Quick Threat Model Creation Using JSON Template
+
+To simplify initial threat model creation, pytm provides a JSON template.
+
+Users can copy `examples/template_tm.json` and modify it to define their threat model.
+This reduces the amount of manual typing required when creating a new threat model.
+
 
 The *colormap* argument, used together with *dfd*, outputs a color-coded DFD where the elements are painted red, yellow or green depending on their risk level (as identified by running the rules).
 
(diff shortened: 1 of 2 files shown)


aryangupta060407-web:
This template complements the existing JSON import functionality by providing an easy starting point for new users.

aryangupta060407-web:
If useful, I can also add a CLI command (e.g., pytm init) to auto-generate this template

izar:
The JSON file seems to be empty, also, how is this supposed to be used?

aryangupta060407-web:
> The JSON file seems to be empty, also, how is this supposed to be used?

Thanks for the feedback!
I’ve updated the JSON template with a concrete example and added documentation explaining how it can be used with pytm’s JSON import functionality (TM.load()).

The goal of this template is to provide a simple starting point so users don’t need to write the full Python code when creating an initial threat model.

Please let me know if you’d like me to adjust the format or add more examples.

izar:
This is a great idea, thanks - but would it be possible for you to use the existing tm.py as the example for the JSON ? It would make things more consistent by offering two views of the same system. 

Also, be advised: we may be deprecating the JSON functionality in favor of the CycloneDX TMBOM quite soon.

aryangupta060407-web:
> This is a great idea, thanks - but would it be possible for you to use the existing tm.py as the example for the JSON ? It would make things more consistent by offering two views of the same system.
> 
> Also, be advised: we may be deprecating the JSON functionality in favor of the CycloneDX TMBOM quite soon.

Thanks a lot for the suggestion, that makes a lot of sense.

Yes, I can update the JSON template to mirror the existing tm.py example so that both represent the same system. That should definitely make the usage clearer and more consistent.

Thanks also for the heads-up about the potential deprecation of JSON in favor of CycloneDX TMBOM. I’ll keep that in mind while updating the example and try to keep the structure as aligned and future-proof as possible.

I’ll push an updated version shortly and would love to get your feedback on it.

aryangupta060407-web:
> This is a great idea, thanks - but would it be possible for you to use the existing tm.py as the example for the JSON ? It would make things more consistent by offering two views of the same system.
> 
> Also, be advised: we may be deprecating the JSON functionality in favor of the CycloneDX TMBOM quite soon.

Thanks for the suggestion!  

I’ve updated the JSON template to mirror the existing tm.py example so that both represent the same system. I kept the JSON version simplified but aligned with the core components and flows from tm.py.  

Please let me know if you’d like me to include more details from tm.py or adjust the structure further.

izar:
Are you sure your addition mirrors the tm.py file ?

aryangupta060407-web:
> Are you sure your addition mirrors the tm.py file ?
I’ve updated the JSON example to more closely mirror the existing tm.py model. 
The goal now is that both representations describe the same system (boundaries, actors, servers, datastores, lambda, data objects, and main dataflows), with the JSON version expressed in a structured and readable way.

While it’s still not a perfect 1:1 mapping of every Python feature (given the current state of JSON support in pytm), it should now be much closer in structure and semantics to tm.py.

If you think certain fields or behaviors should be aligned even further, I’d be happy to refine the example accordingly.

izar:
Why do I feel like I'm talking to an LLM?

PR #317, opened 2026-03-04
Please ignore

Krimdar:
Mistakenly targeted main repo instead of own test repo with merge request. Sorry for the confusion. please ignore.

### DIFF
--- a/docs/reveal.md
+++ b/docs/reveal.md
@@ -46,7 +46,6 @@
 ----
 }
 
-
 ---
 
 ## Actors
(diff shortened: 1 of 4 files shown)


docs/reveal.md
izar, line 145: any specific reason to move away from the table format?

pytm/report_util.py
izar, line 66: this creates dependencies if attributes are added at runtime, is there a more dynamic way to do it?

PR #318, opened 2026-04-01
Add likelihood field to Finding

eirmich:
## Summary
- Adds the `likelihood` string field to the `Finding` class, alongside the existing `cvss` field
- Includes `likelihood` in the serializable fields list in `encode_threat_data`

## How to test
Run the new pytm version with a sample threat model and verified `likelihood` is part of the output:
```
> jq '.findings[0]' /tmp/out.json
{
  "assumption": null,
  "condition": "target.controls.handlesResourceConsumption is False or target.controls.isResilient is False",
  "cvss": "",
  "description": "Flooding",
  "details": "...",
  "example": "Adversary tries to bring a network or service down by flooding it with large amounts of traffic.",
  "id": "1",
  **"likelihood": "High"**,
  "mitigations": "...",
  "response": "",
  "severity": "Medium",
  "target": "React Frontend",
  "threat_id": "DO01"
}
```

## How to validate
- [ ] Verify `likelihood` can be set on a `Finding` object
- [ ] Verify `likelihood` appears in JSON/report output

### DIFF
--- a/pytm/pytm.py
+++ b/pytm/pytm.py
@@ -700,6 +700,7 @@ Can be one of:
 """,
     )
     cvss = varString("", required=False, doc="The CVSS score and/or vector")
+    likelihood = varString("", required=False, doc="Likelihood of the threat")
 
     def __init__(
         self,
@@ -721,6 +722,7 @@ Can be one of:
             "example",
             "references",
             "condition",
+            "likelihood",
         ]
         threat = kwargs.pop("threat", None)
         if threat:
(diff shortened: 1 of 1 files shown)


izar:
Hi @eirmich and thanks for your patch! 
How do you propose the likelihood be set?

eirmich:
Hi @izar, this patch doesn't set but only exposes the defined likelihood as it is already part of the threat (introduced in https://github.com/OWASP/pytm/pull/241).
Let me know if you have any concerns

izar (approved):

Issue #294, opened 2025-10-13
Add docstrings for pytm model entities


fkromer:
The CLI based development workflow using `./tm.py --list-elements` + `./tm.py --describe <element>` is not convenient for threat model creation.

An easier and more efficient workflow to work with model entities (this includes custom model entities) makes use of/depends on docstrings.

E.g. when using google docstring format the [Actor class](https://github.com/OWASP/pytm/blob/master/pytm/pytm.py#L1819-L1831) ([docstring reference](https://google.github.io/styleguide/pyguide.html#384-classes)) + `__init__` method ([google format docstring reference](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods)) could be documented like

```
class Actor(Element):
    """An entity usually initiating actions.

    Attributes:
        port (int): Default TCP port for outgoing data flows.
        protocol (str): Default network protocol for outgoing data flows.
        data (list): List of pytm.Data objects in outgoing data flows.
        inputs (list): List of incoming Dataflows.
        outputs (list): List of outgoing Dataflows.
        isAdmin (bool): Indicates if the actor is an administrator.
    """

    port = varInt(-1, doc="Default TCP port for outgoing data flows")
    protocol = varString("", doc="Default network protocol for outgoing data flows")
    data = varData([], doc="pytm.Data object(s) in outgoing data flows")
    inputs = varElements([], doc="incoming Dataflows")
    outputs = varElements([], doc="outgoing Dataflows")
    isAdmin = varBool(False)

    def __init__(self, name, **kwargs):
        """
        Initialize an Actor object.

        Args:
            name (str): The name of the actor.
            **kwargs: Additional keyword arguments.
                port (int): Default TCP port for outgoing data flows.
                protocol (str): Default network protocol for outgoing data flows.
                data (list): List of pytm.Data objects in outgoing data flows.
                inputs (list): List of incoming Dataflows.
                outputs (list): List of outgoing Dataflows.
                isAdmin (bool): Indicates if the actor is an administrator.
        """
        super().__init__(name, **kwargs)
        TM._actors.append(self)
```

and would result IDE-wise in something like this:

https://github.com/user-attachments/assets/9ede76e1-3689-4906-bd39-b2af7e0333af

fkromer:
Is compatible with #279 .

fkromer:
[...]

PR #320, opened 2026-04-11
Refactor models to Pydantic

NoodlesNZ:
Replaces the custom var descriptor system with Pydantic v2 BaseModel. This simplifies validation and makes it easier to extend. The public API is unchanged, existing threat model scripts should work without modification.

Changes:
  - Split the monolithic pytm.py into per-class modules (element.py, actor.py, asset.py, dataflow.py, tm.py, etc.)
  - All model classes now use Field() for attribute definitions and validators for coercion
  - Added LLM element with supporting threat conditions
  - Removed sqlDump / --sqldump (dropped the pydal dependency)
  - Dropped Python 3.8 support; added 3.12–3.14

Bug fixes:
  - Threat conditions targeting base classes (e.g. Asset) now correctly use isinstance rather than string comparison
  - Finding used as an override stub no longer registers a phantom element in the global state
  - Legacy string assignment to a dataflow's data field was incorrectly classifying data as PUBLIC instead of UNKNOWN
  - Threat.apply() returned None instead of False on type mismatch

Added Tests:
Added tests/test_pydantic_models.py covering the new validation and coercion behaviour (DataSet, field validators, Threat legacy field mapping). All existing tests pass.


pytm/asset.py
```diff
@@ -0,0 +1,163 @@
...
+            from .tm import TM
+            TM._assets.append(self)
+        except ImportError:
+            pass
+
+
+class Lambda(Asset):
+    """A lambda function running in a Function-as-a-Service (FaaS) environment."""
+    
+    onAWS: bool = Field(default=True, description="Is this lambda on AWS")
```
colesmj: Just want to verify - did you intend to recreate this Field in the subclass vs setting the Field inherited?

pytm/dataflow.py
```diff
@@ -0,0 +1,203 @@
...
+    isResponse: bool = Field(default=False, description="Is a response to another data flow")
+    response: Optional['Dataflow'] = Field(
+        default=None,
+        description="Another data flow that is a response to this one"
+    )
+    responseTo: Optional['Dataflow'] = Field(
+        default=None,
+        description="Is a response to this data flow"
+    )
+    srcPort: int = Field(default=-1, description="Source TCP port")
```
colesmj, line 25: A more general comment, not specific to this review but convenient: The srcPort and dstPort descriptions should drop 'TCP' to give them some flexibility to be dataflows for more than just TCP networks.

pytm/pytm.py
```diff
@@ -1,376 +1,61 @@
...
+from .actor import Actor
+from .process import Process, SetOfProcesses
+from .dataflow import Dataflow
+from .boundary import Boundary
+from .tm import TM, UIError
 
 logger = logging.getLogger(__name__)
 
-
-class var(object):
```
colesmj, line 56: Finally....

colesmj (commented):
Thank you for submitting this refactoring of the pytm base code; it was long overdue. The change to pydantic instead of the legacy _var_ mechanism will make for cleaner code and also create some new possibilities for extending pytm.

No threats needed to change to support this refactor (e.g. the condition evals were all "safe" with the checks you added)?

pytm/asset.py
NoodlesNZ: Was just following the original code where `onAWS` defaults to `True` for `Lambda`. To be honest I'm not a fan of using "onAWS" or calling this "Lambda" but I didn't want to change the API for now.

pytm/dataflow.py
NoodlesNZ: I agree, especially when we have a protocol field here too. What are your thoughts about setting the protocol field to default to `TCP`?

NoodlesNZ:
> Thank you for submitting this refactoring of the pytm base code; it was long overdue. The change to pydantic instead of the legacy _var_ mechanism will make for cleaner code and also create some new possibilities for extending pytm.
> 
> No threats needed to change to support this refactor (e.g. the condition evals were all "safe" with the checks you added)?

Thanks for the review and the comments! There are a bunch of changes I'd like to make but I wanted to keep it invisible to most end users. Maybe with more drastic changes this might become part of a 2.0 release?

No threats needed to change for now. I did think about moving the threat library to python code, but there was not a lot to gain other than using enums and classes (to save string matching).

pytm/dataflow.py
colesmj: From a threat perspective that is probably less useful than setting it to HTTP or HTTPS. In truth, any data flow may have layered protocols (e.g. TLS over HTTP over TCP over IPv4); obviously some are more informative for threat analysis than others.

NoodlesNZ:
I've made some changes to add docstrings to help address #294

colesmj (approved):

colesmj (approved):
Good changes, beneficial to all.

NoodlesNZ:
I'm just doing some final clean ups and adding a bit of test coverage. I'll try push the last changes later today. Thank you for the reviews :D

NoodlesNZ:
This should be all ready to go now

colesmj (approved):

PR #321, opened 2026-04-13
Update makefile/dockerfile

NoodlesNZ:
Changes:
- Fix broken URLs in Dockerfile
- Bump base image for Dockerfile
- Update to use poetry to manage dependencies
- Use poetry to run commands in Makefile so that the virtual environment is used
- Add entrypoint.sh to remove complexity/overloading in Makefile

The docker image can now be used like this:
```
docker run --rm -v $(pwd):/work pytm:latest dfd
```
or
```
docker run --rm -v $(pwd):/work pytm:latest dfd other_tm.py
```

This means the docker image can be used without the `Makefile` which makes it more portable

`make` can still be used as per normal, but it removes the complexity of trying to use docker as a shell (which didn't seem to work well)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,29 +1,28 @@
+FROM python:3.14.4-alpine3.23
 
-FROM python:alpine3.20
+WORKDIR /app
+ENTRYPOINT ["/app/entrypoint.sh"]
 
-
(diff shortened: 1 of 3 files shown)


izar (approved):
Thanks!

PR #322, opened 2026-04-15
Some minor type fixes/clean ups

NoodlesNZ:
Applies the valid type fixes from #313, dropping changes that were superseded by the Pydantic refactor in #320.
                                                                                                                                       
Applied:                       
  - flows.py: Fix req_reply return type annotation ((DF, DF) → tuple[DF, DF]) and reply return type (DF → tuple[DF, DF])               
  - template_engine.py: Rename format_field parameter spec → format_spec to match string.Formatter base class signature                
  - pytm.py: Import PydanticUndefined from pydantic_core instead of the private pydantic.fields module                 
  - tm.py: Declare _duplicate_ignored_attrs as a ClassVar rather than monkey-patching it at module level                               
                                                                                                                                       
Dropped (no longer applicable):                                                                                                      
  - varBool string-to-bool coercion — varBool is now a plain type alias; Pydantic handles coercion                                     
  - DataSet.__str__ None guard — DataSet was rewritten in #320 with a _names index that is already safe                                
  - DataSet generic type parameterization — superseded by the new DataSet implementation in base.py

*Tests*
All tests passing, no additional code coverage needed.

### DIFF
--- a/pytm/flows.py
+++ b/pytm/flows.py
@@ -2,7 +2,7 @@ from pytm import Dataflow as DF
 from pytm import Element
 
 
-def req_reply(src: Element, dest: Element, req_name: str, reply_name=None) -> (DF, DF):
+def req_reply(src: Element, dest: Element, req_name: str, reply_name=None) -> tuple[DF, DF]:
     """
     This function creates two datflows where one dataflow is a request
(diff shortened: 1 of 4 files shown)


izar (approved):
LGTM!

Issue #327, opened 2026-04-20
Consider using python objects for defining threats


NoodlesNZ:
The current system of using threats defined in JSON is a little bit hard to use/maintain. The conditions for a threat are written as a string, then eval'ed which is not ideal.

With the pydantic models, it should be easy to define threats as python classes and allow for type hinting, unit tests, linting etc

NoodlesNZ:
Proof of concept: #328

izar:
and that would remove the need for the eval, correct?

NoodlesNZ:
Correct, at least for the built in threats. If you want to allow loading of json files for backwards compatibility then it will have eval

Each threat class has a method called `_check_condition`:
https://github.com/OWASP/pytm/pull/328/changes#diff-c0ca22ec4303e6900ea65a55c408ad3cefdcc1285aa68d68c613c67b19625845R22-R23
This implements the conditional logic of when to apply the threat

izar:
That eval keeps bothering me. There needs to be a more elegant and secure way of doing it short of developing an internal language.

NoodlesNZ:
Fwiw, the `eval()` has always been there, but moved with the pydantic rewrite. This is the original code: https://github.com/OWASP/pytm/blob/ec9a5c1193824972c3ff01026b3de2c291e49b8d/pytm/pytm.py#L672

The eval only exists now for backwards compatibility with the json threat libraries. If we made a breaking change to remove that, then it would no longer be needed.

izar:
I'm thinking if there was ever a case for a breaking change then this is it. @colesmj ?

fkromer:
@NoodlesNZ I like the idea of modelling threats in code using pydantic instead of in JSON.

@izar This would be consistent with the overall pytm „in code“ philosophy and other pydantic related refactorings we already merged. Providing threats as JSON would be possible as well still. Plus… because JSON can be exported from pydantic models issues like https://github.com/OWASP/pytm/issues/237#issuecomment-5285995701 would be prevented by design. This would eliminate the need for testing (because pydantic is tested already) all JSON exporting (+ possible importing) related functionality.

fkromer:
Might relate to and ease to handle https://github.com/OWASP/pytm/issues/152 and https://github.com/OWASP/pytm/issues/14 potentially.

PR #328, opened 2026-04-20
First pass at converting threats to python classes

NoodlesNZ:
Related to #327 here's an implementation of using python classes to define threats. It adds backwards compatibility via `_add_threats_from_json()` in `tm.py`

Created as a draft to show how this is possible and gauge whether this is something that pytm wants to adopt.


raphaelahrens:
Why did you implement the Threats as classes and no just as objects?

Further this change should consider the use case from PR https://github.com/OWASP/pytm/pull/263 , would this only be possible with the JSON format or would this mean that python modules have to be included?

NoodlesNZ:
@raphaelahrens sorry for the delay.

I'm proposing using classes mainly so threats can be subclassed and imported by name. The condition becomes a method you override, so it's ordinary Python that linters, type checkers and tests can see, and each threat has a stable symbol to write a test against (`from pytm.threatlib.inp import INP05`).

Re PR #263, this change would allow for libraries of threats to be imported from other packages

PR #330, opened 2026-04-30
Support Data object decoding in JSON decoder

### LINKED ISSUE #204: Error with data field in input JSON
When using the JSON model format as input to create a report I am getting an error "expecting a list of pytm.Data, item number 0 is a <class 'str'>" (line 213 in the code snippet).
https://github.com/izar/pytm/blob/679ea0df19b7b92e7d8359891d53f7ed794d54a3/pytm/pytm.py#L194-L218
My input JSON for data and flows looks like this:
"flows": [
        {
            "name": "Actor 1 to Actor 2",
            "source": "Actor 1",
            "sink": "Actor 2",
            "order": 1,
            "data": [
                "Data"
            ]
        },{
            "name": "Actor 2 to Actor 3",
            "source": "Actor 2",
            "sink": "Actor 3",
            "description": "Another data flow",
            "data": [
            ]
        }
    ],
    "data": [
        {
            "name": "Data",
            "format": "Text",
            "isPII": true
        }
    ]

Which I believe matches the JSON format when using the JSON output of the CLI tool.  Below is the function which leads to calling varData when it creates the Dataflow object.  It seems like the varData function doesn't deal with a list of data name strings.  [input.json](https://github.com/izar/pytm/blob/679ea0df19b7b92e7d8359891d53f7ed794d54a3/tests/input.json) in the tests folder doesn't have a data field in it.  Since data objects are not in either the boundaries, elements or flows section of the JSON should they be dealt with using their own function e.g. decode_data?
https://github.com/izar/pytm/blob/679ea0df19b7b92e7d8359891d53f7ed794d54a3/pytm/json.py#L92-L107

mlevogiannis:
This PR extends the JSON decoder to support the Data type. It enables the decoder to load Data objects and implements reference resolution within DataFlow objects to point to the newly loaded Data objects.

The corresponding test case has also been updated to account for the new functionality.

Fixes #204.

### DIFF
--- a/pytm/json.py
+++ b/pytm/json.py
@@ -24,24 +24,17 @@ _ELEMENT_CLASSES = {
 
 def loads(s):
     """Load a TM object from a JSON string *s*."""
-    result = json.loads(s, object_hook=decode)
-    if not isinstance(result, TM):
-        raise ValueError("Failed to decode JSON input as TM")
-    return result
+    result = json.loads(s)
+    return decode(result)
 
 
 def load(fp):
     """Load a TM object from an open file containing JSON."""
-    result = json.load(fp, object_hook=decode)
-    if not isinstance(result, TM):
-        raise ValueError("Failed to decode JSON input as TM")
-    return result
+    result = json.load(fp)
+    return decode(result)
 
 
 def decode(data):
-    if "elements" not in data and "flows" not in data and "boundaries" not in data:
-        return data
-
     boundaries = decode_boundaries(data.pop("boundaries", []))
     elements = decode_elements(data.pop("elements", []), boundaries)
     decode_flows(data.pop("flows", []), elements)
--- a/pytm/json.py
+++ b/pytm/json.py
@@ -1,13 +1,15 @@
 import json
 
 from .tm import TM
+from .base import DataSet
 from .boundary import Boundary
+from .data import Data
 from .dataflow import Dataflow
 from .asset import Asset, Server, ExternalEntity, Lambda, LLM
 from .datastore import Datastore
 from .actor import Actor
 from .process import Process, SetOfProcesses
-from .enums import Action
+from .enums import Action, Classification, Lifetime
 
 _ELEMENT_CLASSES = {
     "Asset": Asset,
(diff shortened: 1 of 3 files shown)


pytm/json.py
```diff
@@ -24,33 +26,27 @@
...
-        raise ValueError("Failed to decode JSON input as TM")
-    return result
+    result = json.load(fp)
+    return decode(result)
 
 
-def decode(data):
-    if "elements" not in data and "flows" not in data and "boundaries" not in data:
-        return data
+def decode(flat):
```
NoodlesNZ: Could we rename `decode()` to make it an internal method? E.g. `def _decode(flat):`

As this method now makes changes to the object `flat`, it could cause problems if `decode()` was called directly. It's only referenced within this file, so it doesn't have any impact on anything else, but will stop problems in the future.

pytm/json.py
```diff
@@ -74,6 +70,30 @@ def decode_boundaries(flat):
...
+    data = {}
+    for i, e in enumerate(flat):
+        name = e.pop("name", None)
+        if name is None:
+            raise ValueError(f"name property missing in data {i}")
+
+        classification_name = e.pop("classification", None)
+        if classification_name:
+            try:
+                e["classification"] = getattr(Classification, classification_name)
```
NoodlesNZ: can this be changed to:
```
e["classification"] = Classification(classification_name)
```
to make it consistent with how `Lifetime` is being used below?

Making this change also means that the try/catch is not needed as `Classification`  will raise an error if a bad value is passed

NoodlesNZ (commented):
Looks good, just a couple of minor improvements. We may also want to make sure that we have test coverage for classification/lifetime

pytm/json.py
mlevogiannis: Done.

pytm/json.py
mlevogiannis: The call syntax (e.g. `Classification(classification_name)`) requires the member's assigned value, which is this case is an integer. But we can use the subscript notation (e.g. `Classification[classification_name]`)  which resolves the member by its key name. I made the change.

For consistency, I changed Lifetime resolution to use the subscript notation, too. But this has one important consequence: The expected input value is now the member's key name (as in Classification) rather than the assigned string value (as it was in the previous version when the call syntax was used).

mlevogiannis: For consistency, we may also change Action resolution (function `_decode`, line 48) to use the subscript notation. This will not be a breaking change, as its members' keys and values currently match. I can make this change, too, if you agree.

mlevogiannis:
> Looks good, just a couple of minor improvements. We may also want to make sure that we have test coverage for classification/lifetime

I have updated the test case to take into account the classification and lifetime, too.

NoodlesNZ:
> > Looks good, just a couple of minor improvements. We may also want to make sure that we have test coverage for classification/lifetime
> 
> I have updated the test case to take into account the classification and lifetime, too.

thank you! LGTM

izar:
On NoodlesNZ review :)

izar (approved):

PR #305, opened 2025-11-22
Migrate project towards cutting edge 2025 tooling and methodology

fkromer:
- restructure project to enable clean [Python library](https://docs.astral.sh/uv/concepts/projects/init/#libraries) builds
- replace [`poetry`](https://github.com/python-poetry/poetry) with [`uv`](https://github.com/astral-sh/uv)
  - enable usage of [`uv` workspaces](https://docs.astral.sh/uv/concepts/projects/workspaces/)
  - add [CycloneDX](https://github.com/CycloneDX) sbom output
- `pyproject.toml`
  - added state of the art dev tools (test framework, static analysis, type checking)
  - single source of package build
  - single source of Python dependencies (`requirements.txt` files generated from it) 
- fix and extend contribution docs
- remove obsolete files
- (backwards compatible Dockerfile adjustments... Dockerfile setup still not working, adding a [devcontainers](https://github.com/devcontainers) Dockerfile would make more sense anyways I guess... Dockerfile for report and diagram generation are usually maintained by `pytm` users due to "own" container security analysis tools)
- (Makefile setup not tested... is obsolete as soon as devcontainer Dockerfile is available anyways I guess)

Issue #302, opened 2025-11-19
Strategy for decoupling models, rules from runtime


fkromer:
While one of pytm biggest strength is the "threat model as code" concept atm it's not easily possible to decouple customer specific rules, model libs, etc. from the pytm "runtime" (report generation, diagram generation). This topic relates to https://github.com/OWASP/pytm/issues/295#issuecomment-3399078011 and https://github.com/OWASP/pytm/issues/19 as well potentially. E.g. ideally all the "models as code" could be handled separately from the pytm CLI application capabilities. This would significantly simplify enable people to integrate pytm into CI workflows as well as to scale pytm usage in general. 

@izar What are your longer term goals related to this topic?

izar:
We do have at least a midterm solution in the form of user-loadable rule files. It is not ideal but it works. 
One thing to keep in mind is that not every user of pytm is a security person who will develop rules. Those we have seen in the past were not shy of adding to the existing json file. Most users are either security practitioners that just want something that works, or developers who just want to use the existing rules. So this would be a 80-20 split between end users and the difficulty of implementing a solution. As a big proponent of good enough is better than perfect, I believe we need to introduce an alternative that a) won't break things for most people, b) will be simple to migrate to for those who it breaks.

Considering the other thoughts around moving to data classes, introducing pydantic, etc. - we may need to stabilize some of those changes before we move into this one.

fkromer:
> We do have at least a midterm solution in the form of user-loadable rule files. It is not ideal but it works.

A mid-term solution I could think of would be to provide users help about possible current solutions. E.g. using a patched version of `pytm` using `uv` instead of `poetry` (like in this experimental branch https://github.com/meteocontrol/pytm/tree/meteocontrol-uv) enables usage of [uv workspaces](https://docs.astral.sh/uv/concepts/projects/workspaces/) which would be one mid-term solution.

> One thing to keep in mind is that not every user of pytm is a security person who will develop rules. Those we have seen in the past were not shy of adding to the existing json file. Most users are either security practitioners that just want something that works, or developers who just want to use the existing rules. So this would be a 80-20 split between end users and the difficulty of implementing a solution. As a big proponent of good enough is better than perfect, I believe we need to introduce an alternative that a) won't break things for most people, b) will be simple to migrate to for those who it breaks.

I absolutely agree.

> Considering the other thoughts around moving to data classes, introducing pydantic, etc. - we may need to stabilize some of those changes before we move into this one.

[...]

PR #331, opened 2026-05-17
Modernize python package setup

Alicipy:
While starting with pytm, I saw a few things regarding package management that could use improvement, especially as the Python folks use pure pip or uv currently, and maybe wondering why the commands don't work.

This PR includes migration to the default pyproject.toml keys which poetry in version 2 also understands, so nothing breaks.  As pip can handle these as well, we can remove a few other files which existed for compatibility reasons (?) as well.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -32,7 +32,7 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install Poetry
-        run: pip install poetry
+        run: pip install "poetry>=2.1"
       - name: Update poetry.lock
         run: poetry lock 
       - name: Install dependencies
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,18 +22,20 @@ dependencies = [
 [project.urls]
 Homepage = "https://github.com/OWASP/pytm"
 
+[dependency-groups]
+dev = [
+    "pytest>=8.3.5,<10.0.0",
+    "black>=25.9,<27.0",
+    "pdoc3>=0.11.6,<0.12.0",
+    "ruff>=0.15.11,<0.16.0",
+]
+
 [tool.poetry]
 packages = [{include = "pytm"}]
 include = [
     "pytm/images/*.png",
     "pytm/threatlib/threats.json",
 ]
-
-[tool.poetry.group.dev.dependencies]
-pytest = ">=8.3.5,<10.0.0"
-black = ">=25.9,<27.0"
-pdoc3 = "^0.11.6"
-ruff = "^0.15.11"
 
 [tool.ruff]
 target-version = "py311"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,19 +1,13 @@
-[tool.poetry]
+[project]
 name = "pytm"
 version = "1.4.0"
 description = "A Pythonic framework for threat modeling"
-authors = ["pytm Team <please_use_github_issues@nowhere.com>"]
-license = "MIT License"
+authors = [{ name = "pytm Team", email = "please_use_github_issues@nowhere.com" }]
+license = "MIT"
 readme = "README.md"
-homepage = "https://github.com/OWASP/pytm"
-packages = [{include = "pytm"}]
-include = [
-    "pytm/images/*.png",
-    "pytm/threatlib/threats.json",
-]
+requires-python = ">=3.11,<3.15"
 classifiers = [
     "Programming Language :: Python :: 3",
-    "License :: OSI Approved :: MIT License",
     "Operating System :: OS Independent",
     "Development Status :: 5 - Production/Stable",
     "Environment :: Console",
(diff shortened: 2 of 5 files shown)


izar (approved):

izar:
Hi and thanks for the PR! Can you work with @fkromer on #305 ? Otherwise I believe we'll have some conflicts here.

Alicipy:
Yes of course, I would love to! How do you suggest to proceed? Shall we merge this and redo the other one, or vice versa? This change does work in both, poetry and uv; as it depends only on PEP-specified behaviour right now.
I am also happy to close this one, and support on the other one fully.

fkromer:
@Alicipy You missed the github ci related changes I guess. Some other notes: gitbook support seems to be deprecated. Using a docs dir is the defacto standard compatible with mkdocs and other Python ecosystem static documentation generators. I‘d recommmend to have a look into https://github.com/OWASP/pytm/pull/305/changes#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711 (and the overall PR) which addresses those topics already.

pyproject.toml
```diff
@@ -21,16 +17,21 @@ classifiers = [
...
-[tool.poetry.group.dev.dependencies]
-pytest = "^8.3.5"
-black = "^25.9.0"
-pdoc3 = "^0.11.6"
-ruff = "^0.15.11"
+[dependency-groups]
+dev = [
+    "pytest>=8.3.5,<9.0.0",
+    "black>=25.9.0,<26.0.0",
+    "pdoc3>=0.11.6,<0.12.0",
```
fkromer: All modern projects use mkdocstings instead. E.g. FastAPI https://github.com/fastapi/fastapi/blob/master/pyproject.toml#L140 which is a good reference.

Alicipy:
I have split the docs out, and integrated the CI work from @fkromer.

I kept it tool agnostic for now. While I personally also like uv, I do not want to break any workflow that is preferable by the maintainers currently, but would help migrating further if wanted :)

Thanks for the work on pytm so far!

fkromer:
@Alicipy Providing better backwards compatibility is a good idea 😁 the rather oldschool pdoc3 is docs only… not really relevant to me.

izar:
Just a thought - I think it would make sense to remove the poetry stuff from CONTRIBUTING.md - we don't need to give instructions for another package that may diverge over time.

.github/workflows/main.yml
fkromer, line 24: From a functional point of view it would make sense to migrate yet some other change from my initial MR https://github.com/OWASP/pytm/pull/305/changes#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R24: Would make sense to support all `python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]` not end of life yet: https://endoflife.date/python

pyproject.toml
```diff
@@ -1,30 +1,26 @@
-[tool.poetry]
+[project]
 name = "pytm"
 version = "1.4.0"
 description = "A Pythonic framework for threat modeling"
-authors = ["pytm Team <please_use_github_issues@nowhere.com>"]
-license = "MIT License"
+authors = [{ name = "pytm Team", email = "please_use_github_issues@nowhere.com" }]
```
fkromer, line 5: Would make sense to give all contributors contribution by name here as well I guess: https://github.com/OWASP/pytm/pull/305/changes#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711

fkromer (commented):
Formatting does not really need `black` because `ruff` provides a black compatible formatter https://docs.astral.sh/ruff/formatter/ and linter https://docs.astral.sh/ruff/faq/#is-the-ruff-linter-compatible-with-black out of the box.

fkromer:
@Alicipy Hint: Might relate to https://github.com/OWASP/pytm/issues/302 .

.github/workflows/main.yml
Alicipy: The current main branch supports only >= 3.11, so this would not work. Given that 3.10 is supported for only 2 more month, I would not touch this.

pyproject.toml
Alicipy: I agree, but I would put that in a different PR, as this should only touch the project setup.

Alicipy:
> Just a thought - I think it would make sense to remove the poetry stuff from CONTRIBUTING.md - we don't need to give instructions for another package that may diverge over time.

I hope I removed all references.

Alicipy:
> @Alicipy Hint: Might relate to #302 .

Read through it, but I would to this separately.

Alicipy:
> Formatting does not really need `black` because `ruff` provides a black compatible formatter https://docs.astral.sh/ruff/formatter/ and linter https://docs.astral.sh/ruff/faq/#is-the-ruff-linter-compatible-with-black out of the box.

I  agree, but adding 'ruff' in this PR would reformat a few files, I would so that in a clean new PR instead.

pyproject.toml
Alicipy: I would rather not, and I think there is a better home for it. This is PyPi metadata and ends up in the wheels Metadata on the projecdt page. If at all, then only the main maintainer should have it's mail here.
If @izra wants it, he can put it in, but I would say the CONTRIBUTOR.md file is the beter place for it.

Alicipy:
@izar Hope I fixed everything properly, ready for re-review / merge

pyproject.toml
fkromer: Makes sense… Note: mkdocstrings is deprecated by Zensical in the meanwhile. Relates to https://github.com/OWASP/pytm/issues/352 .

fkromer:
Thanks for the work put into here @Alicipy . LGTM. @izar already approved but might want to re-approve again probably.

Alicipy:
Thanks for taking the time to review and improve it @fkromer :)

izar (approved):

PR #332, opened 2026-05-22
Fixing documentation to match new Elements

izar:
Added LLM and Agent, to tm.py and readme.


colesmj (approved):
Changes look good. At some point it would be worthwhile revisiting LLM specifically (vs GenAI Model, perhaps) but not a blocker for this PR.

fkromer:
@izar @colesmj Not sure if this really relates to this MR... just a general note:

In general the risk related to LLM and agent usage heavily depends on where the data processing happens. E.g. if local or self-hosted LLMs are used technologically there is way less risk for all of https://github.com/OWASP/pytm/pull/332/changes#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R552-R565 (despite of hard-coded secret which might end up in git histories and code). The degree of risk remaining depends on the backend/infrastructure used of course.

izar:
@fkromer as much everything, the devil is in the details :D 
simply using a local LLM lowers one risk of data leak, for sure. But that assumes programatic use? What if you're using Claude Code with a model off of ollama, but then your skills do the leak for you? 
Guess it is time to add more attributes to the new elements!

fkromer:
@izar Yes, this hint is better placed as a hint in https://github.com/izar/tm_skills I guess.

PR #337, opened 2026-07-06
fix: remove dead builder methods and CLI output typo

izar:
## Summary

- **Dead methods deleted from `_FlowDefaultsBuilder`**: `_color`, `display_name`, `_label`, `_shape`, `_safeset` (lines 414–433 in the pre-fix file). These were orphaned copies of `Element`-level methods accidentally left inside the `@dataclass` during the Pydantic migration. They were never called on the builder — the real implementations live on the element classes. Additionally, `_label` called `wrap(...)` which is not imported in `pytm.py` (it's imported in `element.py`), meaning any invocation would have raised `NameError`. Deletion resolves the latent bug.
- **`_list_elements` typo fixed**: `print("\nAtributes:")` → `print("\nAttributes:")`. No test asserted on the misspelled string.
- **`Element` added to `_CLASS_REGISTRY`**: `_describe_classes` iterated only subclasses of `Element`, not `Element` itself, so `--describe Element` errored with "No such class to describe: Element". One-line fix restores the documented CLI behaviour.

## Validation

```
python3 tm.py --list-elements   # "Attributes:" printed correctly; no NameError
python3 tm.py --describe Element  # works, shows current Pydantic fields
uv run pytest tests/ -x -q      # 243 passed
git diff --stat HEAD~1           # pytm/pytm.py | 23 +++...--- (no README touches)
```


izar:
I'm merging as a reviewer of AI-generated code...

PR #347, opened 2026-08-11
Poetry updates

izar:
FIxing poetry in the Github actions.


izar:
This one bypassed review because it is infra-structure, not functionality.

PR #357, opened 2026-08-15
docs: fix Devbox nix config typo

rajanpanth:
## Summary

Fixes a duplicated word in the Devbox setup instructions: `file file` -> `file`.

## Validation

- `git diff --check`
- `rg -n "file file" README.md docs CONTRIBUTING.md` returns no matches

AI assistance disclosure: I used AI assistance to identify and verify this documentation-only typo fix, and I reviewed the final diff.

PR #356, opened 2026-08-15
docs: fix JSON example typo

rajanpanth:
## Summary

Fixes a typo in the README JSON import section: `exmaple` -> `example`.

## Validation

- `git diff --check`
- `rg -n "exmaple" README.md docs pytm tests` returns no matches

AI assistance disclosure: I used AI assistance to identify and verify this documentation-only typo fix, and I reviewed the final diff.

PR #355, opened 2026-08-15
docs: fix report template reference

rajanpanth:
## Summary

Updates stale `docs/template.md` references to the existing `docs/basic_template.md` report template in:

- CLI help text
- README usage output
- sample model `sourceFiles`

## Validation

- `git diff --check`
- `python -m py_compile tm.py pytm/pytm.py`
- `python tm.py --help` shows `docs/basic_template.md`

AI assistance disclosure: I used AI assistance to identify and verify this documentation/source-reference fix, and I reviewed the final diff.

PR #353, opened 2026-08-15
docs: fix RevealMD README references

rajanpanth:
## Summary

Fixes two README issues in the slides section:

- corrects malformed Markdown for the RevealMD link so it renders as a clickable link
- updates the template reference from nonexistent `docs/revealjs.md` to existing `docs/reveal.md`

## Validation

- `git diff --check`
- confirmed `docs/reveal.md` exists and `docs/revealjs.md` does not
- confirmed no `README.md` matches for `revealjs.md` or malformed `(RevealMD)[...]` syntax

AI assistance disclosure: I used AI assistance to identify and verify this documentation-only change, and I reviewed the final diff.

PR #360, opened 2026-08-15
fix: escape boundary labels in DFD output

### LINKED ISSUE #278: Support ">" in Boundaries

rajanpanth:
## Summary

- Escapes boundary names before inserting them into Graphviz HTML-like labels
- Adds a regression test for a boundary name containing `>`

Closes #278.

## Testing

- Direct reproduction check:
  - generated a DFD with `Boundary("Client > API")`
  - verified output contains `Client &gt; API` and not the unescaped label
- Not run: full `pytest` suite is unavailable because `pytest` is not installed in the current Python environment.

## Notes

AI assistance was used to help prepare this change.


NoodlesNZ:
LGTM, although you need to fix your AI environment if you're going to suggest changes like this. It's trivial to get `pytest` working, you can run it via poetry easily. I've validated that it passes and when the Github CI kicks off, it will also validate this.

rajanpanth:
Thanks for going through these. Grouping the follow-up into one comment rather than five.

These five are all green and mergeable, and have an LGTM from you:

- #353 — docs: fix RevealMD README references
- #355 — docs: fix report template reference
- #356 — docs: fix JSON example typo
- #357 — docs: fix Devbox nix config typo
- #360 — fix: escape boundary labels in DFD output (this one)

They still show `REVIEW_REQUIRED`, because the LGTMs were posted as comments rather than as approving reviews, so the required-review gate hasn't cleared. If you're able to convert one into an approving review they should be unblocked; if merging sits with someone else, no rush — just flagging that they're ready rather than waiting on me for anything.

On your point in #360 about the environment: fair, and I've fixed it. I have the suite running locally now — `pytest tests/` gives 243 passed, 4 failed, and the 4 (`test_dfd`, `test_dfd_colormap`, `test_dfd_duplicates_ignore`, `test_multilevel_dfd`) fail identically on `master`; they're Windows path-separator diffs in the DFD output, not related to any of these changes.

PR #362, opened 2026-08-15
fix: reject unknown names in threat conditions

rajanpanth:
## Summary
- reject unknown names while validating threat condition expressions
- preserve names introduced by list/generator comprehensions
- add regression coverage for unknown names and comprehension variables

## Testing
- `py -3.12 -m compileall pytm\threat.py tests\test_pydantic_models.py`
- `git diff --check`
- `py -3.12 -m pytest tests\test_pydantic_models.py::TestConditionValidator -q` in a temporary local venv


NoodlesNZ:
Looks fine to me, although should include some additional changes:

`_build_eval_globals` should include an entry for `LLM`. While this doesn't break anything right now, if someone was writing threats using this element then it would error out.

We should also include a mention in `CHANGELOG` when we next release as this is a breaking change and may catch out legacy users.

rajanpanth:
Both points addressed.

`LLM` is now in `_build_eval_globals` (`2e40cf1`). It was the only element class missing from the map, so conditions referencing it raised `NameError` even though eight shipped threats target the element.

The CHANGELOG entry is reworked in `97a4fbe`. I'd originally filed the `LLM` addition itself under BREAKING, which was wrong — the breaking part is the strict rejection of unknown names. That's now its own entry describing what legacy users will actually hit (fail-fast when the `Threat` is constructed, instead of `NameError` later during evaluation) and how to resolve it, with the `LLM` fix moved under Fixed.

On the environment — fair. I have the suite running locally now. `pytest tests/` gives 243 passed, 4 failed: `test_dfd`, `test_dfd_colormap`, `test_dfd_duplicates_ignore` and `test_multilevel_dfd`, all Windows path-separator diffs (`pytm/images/...` vs `pytm\images\...`). They fail identically on `master`, so they're not from this PR.

One thing I hit that may be worth its own issue: `tests/test_pytmfunc.py` can't be collected at all on Windows without `PYTHONUTF8=1`, because line 38 opens `threats.json` without an explicit encoding and cp1252 fails on it.

izar:
Are we talking to a person or to a LLM ?

rajanpanth:
A person. Fair question though. I use an AI assistant heavily, including for that last comment. The decisions and the review are mine, and I stand behind the change.

On substance: `pytm.LLM` was the only element class missing from `_build_eval_globals`, so conditions referencing it raised `NameError` while eight shipped threats target it. The breaking part isn't that addition. It's that unknown names now fail when the `Threat` is constructed rather than later during evaluation, which is what the CHANGELOG says now.

Happy to answer anything directly. If you'd rather I slow the PR rate here, say so.

PR #365, opened 2026-08-20
Switch to ruff

Alicipy:
As already mentioned in #331, this PR will switch 'black' fully to ruff.

There was a configuration for ruff already, but black was the main formater used before. This change removes black, reformats all files, fixes some linting errors, and enforces the rules in CI as well, so no drift happens now.


NoodlesNZ:
LGTM

